PR #24, opened 2016-12-22
Bugfix mobile 671 fix crash

EarlOfEgo:
## Information
When the activity goes to the background in the review/analyse screen and the system "kills" it(due to resource allocations), or "Don't keep activities" in the developer options are activate, the screen library crashes. This PR fixes it, since the fragment was only created and not retained when `Activity#onCreate()` was called with a saved instance.
## How to test
Go to the review screen and bring the app to the back- and then foreground. Enable/disable the developer option "Don't keep Activities".


a-szotyori:
Fuuuu merged into master

a-szotyori:
Reverted master with a hard reset to erase this mistake.

PR #27, opened 2017-01-16
Feature mobile 689 remove 4 3 requirement

a-szotyori:
## Dropped 4:3 requirement and limitation
We don't check, if the camera has a picture size larger than ~8MP with a 4:3 aspect ratio. We only check that there is a size larger than ~8MP.

The camera uses the largest resolution and the preview resolution is selected to match the picture size's aspect ratio.

## How to test
Run the Screen API or the Component API example app. The Samsung Galaxy S6 has a 16:9 largest resolution and you should see bars on the left and right in the Camera Screen.

Run the tests in `ginivision/src/androidTest` and `ginivision/src/test`.

[MOBILE-689](https://tickets.i.gini.net/issue/MOBILE-689)

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
@@ -1,5 +1,6 @@
 package net.gini.android.vision.internal.camera.api;
 
+import static net.gini.android.vision.internal.camera.api.Util.getLargestSameAspectRatioSize;
 import static net.gini.android.vision.internal.camera.api.Util.getLargestSize;
 
 import android.app.Activity;
@@ -336,16 +337,6 @@ public class CameraController implements CameraInterface {
 
         Camera.Parameters params = mCamera.getParameters();
 
-        List<Camera.Size> previewSizes = params.getSupportedPreviewSizes();
-        Size previewSize = getLargestSize(previewSizes);
-        if (previewSize != null) {
-            mPreviewSize = previewSize;
-            params.setPreviewSize(mPreviewSize.width, mPreviewSize.height);
-            LOG.debug("Preview size ({}, {})", mPreviewSize.width, mPreviewSize.height);
-        } else {
-            LOG.warn("No suitable preview size found");
-        }
-
         List<Camera.Size> pictureSizes = params.getSupportedPictureSizes();
         Size pictureSize = getLargestSize(pictureSizes);
         if (pictureSize != null) {
@@ -354,6 +345,16 @@ public class CameraController implements CameraInterface {
             LOG.debug("Picture size ({}, {})", mPictureSize.width, mPictureSize.height);
         } else {
             LOG.warn("No suitable picture size found");
+        }
+
+        List<Camera.Size> previewSizes = params.getSupportedPreviewSizes();
+        Size previewSize = getLargestSameAspectRatioSize(previewSizes, mPictureSize);
+        if (previewSize != null) {
+            mPreviewSize = previewSize;
+            params.setPreviewSize(mPreviewSize.width, mPreviewSize.height);
+            LOG.debug("Preview size ({}, {})", mPreviewSize.width, mPreviewSize.height);
+        } else {
+            LOG.warn("No suitable preview size found");
         }
 
         if (params.getSupportedFocusModes().contains(Camera.Parameters.FOCUS_MODE_CONTINUOUS_PICTURE)) {
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
@@ -1,7 +1,7 @@
 package net.gini.android.vision.internal.camera.api;
 
-import static net.gini.android.vision.internal.camera.api.Util.getLargestSizeWithSameAspectRatio;
-import static net.gini.android.vision.internal.camera.api.Util.getLargestSize;
+import static net.gini.android.vision.internal.camera.api.SizeSelectionHelper.getLargestSizeWithSameAspectRatio;
+import static net.gini.android.vision.internal.camera.api.SizeSelectionHelper.getLargestSize;
 
 import android.app.Activity;
 import android.graphics.Matrix;
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
@@ -336,7 +336,16 @@ public class CameraController implements CameraInterface {
         }
 
         Camera.Parameters params = mCamera.getParameters();
+        selectPictureSize(params);
+        selectPreviewSize(params);
+        selectFocusMode(params);
+        selectFlashMode(params);
+        mCamera.setParameters(params);
 
+        setCameraDisplayOrientation(activity, mCamera);
+    }
+
+    private void selectPictureSize(final Camera.Parameters params) {
         List<Camera.Size> pictureSizes = params.getSupportedPictureSizes();
         Size pictureSize = getLargestSize(pictureSizes);
         if (pictureSize != null) {
@@ -346,7 +355,9 @@ public class CameraController implements CameraInterface {
         } else {
             LOG.warn("No suitable picture size found");
         }
+    }
 
+    private void selectPreviewSize(final Camera.Parameters params) {
         List<Camera.Size> previewSizes = params.getSupportedPreviewSizes();
         Size previewSize = getLargestSameAspectRatioSize(previewSizes, mPictureSize);
         if (previewSize != null) {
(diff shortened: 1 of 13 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
```diff
@@ -336,24 +337,24 @@ private void configureCamera(Activity activity) {
...
             mPictureSize = pictureSize;
             params.setPictureSize(mPictureSize.width, mPictureSize.height);
             LOG.debug("Picture size ({}, {})", mPictureSize.width, mPictureSize.height);
         } else {
-            LOG.warn("No 4:3 picture size found");
+            LOG.warn("No suitable picture size found");
+        }
+
+        List<Camera.Size> previewSizes = params.getSupportedPreviewSizes();
+        Size previewSize = getLargestSameAspectRatioSize(previewSizes, mPictureSize);
```
nsobadjiev: Maybe not the best name? What about `getLargestSizeWithRatio`?

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
```diff
@@ -336,24 +337,24 @@ private void configureCamera(Activity activity) {
...
+        if (previewSize != null) {
+            mPreviewSize = previewSize;
+            params.setPreviewSize(mPreviewSize.width, mPreviewSize.height);
+            LOG.debug("Preview size ({}, {})", mPreviewSize.width, mPreviewSize.height);
+        } else {
+            LOG.warn("No suitable preview size found");
```
nsobadjiev: Without the 4:3 limitation, going to this else would be a serious problem. It means the camera has no possible resolutions e.g. it doesn't work at all.

Will an exception be thrown elsewhere because of this? Because just a log message is probably not enough.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
```diff
@@ -336,24 +337,24 @@ private void configureCamera(Activity activity) {
...
-        Size previewSize = getLargestFourThreeRatioSize(previewSizes);
-        if (previewSize != null) {
-            mPreviewSize = previewSize;
-            params.setPreviewSize(mPreviewSize.width, mPreviewSize.height);
-            LOG.debug("Preview size ({}, {})", mPreviewSize.width, mPreviewSize.height);
-        } else {
-            LOG.warn("No 4:3 preview size found");
-        }
-
         List<Camera.Size> pictureSizes = params.getSupportedPictureSizes();
```
nsobadjiev: IMO, the `configureCamera` method is starting to get overburdened. It does several things:
* size selection
* Focus
* Flash
* Settings configuration

I'd extract the size selection in a separate method, so at least it doesn't expose the selection algorithm.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/Util.java
```diff
@@ -14,15 +15,45 @@
 public final class Util {
```
nsobadjiev: This `Util` class is very Aspect Ratio and sizes orientated. Perhaps it's not really utils, but `SizeUtils`, `AspectRatios` or `ResolutionSelection`?

ginivision/src/test/java/net/gini/android/vision/requirements/CameraResolutionRequirementTest.java
```diff
@@ -12,49 +12,44 @@
...
 
         CameraResolutionRequirement requirement = new CameraResolutionRequirement(cameraHolder);
 
         assertThat(requirement.check().isFulfilled()).isFalse();
+        assertThat(requirement.check().getDetails()).isEqualTo("Largest camera picture resolution is lower than 8MP");
     }
 
     @Test
-    public void should_reportFulfilled_ifPreviewSize_andPictureSize_with4to3AspectRatio_andPictureSize_isLargerThan8MP() {
+    public void should_reportFulfilled_ifPreviewSize_andPictureSize_isLargerThan8MP() {
```
nsobadjiev: Personally, I don't like the naming convention for test names. I know it's in your code style guide, though, so it's ok.
[Android's guide](http://source.android.com/source/code-style.html#javatests-style-rules) uses kind of the same thing with the underscores. However, they start names with "test".
_No action for this, just expression an opinion_

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
a-szotyori: Good point, I'll change it.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
a-szotyori: Yes, that is true, but it should never happen (no, really). The CameraApi states [here](https://developer.android.com/reference/android/hardware/Camera.Parameters.html#getSupportedPictureSizes()) and [here](https://developer.android.com/reference/android/hardware/Camera.Parameters.html#getSupportedPreviewSizes()) that there will be least one returned resolution. If there is really no size then there is something seriously wrong with the device.

We also have our requirement checks, which our clients use to verify that the device meets the hardware requirements.

ginivision/src/test/java/net/gini/android/vision/requirements/CameraResolutionRequirementTest.java
a-szotyori: JUnit3 required test methods to start with `test`. It is fine, if you don't like the naming convention. There are a lot of ways of writing it and I personally find this easier to read than camel case only or underscore only.

The future will be [string function identifiers](http://tengio.com/blog/more-readable-tests-with-kotlin/) (maybe @EarlOfEgo knows what's the correct term). Another reason for using Kotlin.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/Util.java
a-szotyori: After reading and presenting Clean Code I should start refactoring code my younger self vomited out. 😁

a-szotyori:
@nsobadjiev Implemented the review notes. If it's fine for you, please merge 😉

ginivision/src/test/java/net/gini/android/vision/requirements/CameraResolutionRequirementTest.java
EarlOfEgo: 🤔 actually I'm not sure if this is the way how this should be used. I know that this is the way to call functions from Java code that are keywords in Kotlin. To declare a function like this looks crazy to me, also how would I call such function? But I'll investigate into this kind of stuff.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
```diff
@@ -335,45 +336,56 @@ private void configureCamera(Activity activity) {
         }
 
         Camera.Parameters params = mCamera.getParameters();
+        selectPictureSize(params);
+        selectPreviewSize(params);
+        selectFocusMode(params);
+        selectFlashMode(params);
```
a-szotyori: Good point. I thought about this and what I could propose is something like:
```
// The CameraConfiguration would be an interface, this would allow to change the configuration in it's implementation
newParams = cameraConfiguration.apply(params);
mCamera.setParameters(newParams);
```
In the `apply()` method we can then change the parameters as we see fit.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/SizeSelectionHelper.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @exclude 
+ */
+public final class SizeSelectionHelper {
+
+    @Nullable
+    public static Size getLargestSize(@NonNull List<Camera.Size> sizes) {
```
a-szotyori: Because `new Size(0, 0)` for initializing the preview and picture sizes in the `CameraController`. We could also work with `Camera.Size`, but since it's constructor is private, it's good I think, if we have our own class, even though it does nothing special.

ginivision/src/test/java/net/gini/android/vision/requirements/CameraResolutionRequirementTest.java
a-szotyori: Let the testing framework call it 😉 . For test writers readability and clarity is king so I kinda like this approach 😃 .

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/SizeSelectionHelper.java
EarlOfEgo: true, the constructor is private, had troubles there as well some time.

PR #31, opened 2017-01-25
Feature auto focus run if no cont focus

a-szotyori:
If the device's camera doesn't support continuous focus mode we want to make sure, that there was at least one auto-focus run. For this we do an auto-focus run before taking a picture when the user taps the trigger button.

### How to test
_I will add a test for verifying this behaviour with a mock Camera instance. But until then:_

Run the `ginivision/src/test/(...)/internal/camera/api/CameraParametersHelperTest`.
We don't have devices without cont. focus so what can be tested is that it works as before on our devices.
We will ask Katharina for her S3 Mini which doesn't have cont. focus to verify the behaviour.

### Merging
Conflicts, will be merged locally.

### Ticket 
No ticket.

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
@@ -1,5 +1,10 @@
 package net.gini.android.vision.internal.camera.api;
 
+import static net.gini.android.vision.internal.camera.api.CameraParametersHelper
+        .isFlashModeSupported;
+import static net.gini.android.vision.internal.camera.api.CameraParametersHelper
+        .isFocusModeSupported;
+import static net.gini.android.vision.internal.camera.api.CameraParametersHelper.isUsingFocusMode;
 import static net.gini.android.vision.internal.camera.api.SizeSelectionHelper
         .getLargestSizeWithSameAspectRatio;
 import static net.gini.android.vision.internal.camera.api.SizeSelectionHelper.getLargestSize;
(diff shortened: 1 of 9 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraParametersHelper.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @exclude
+ */
+final class CameraParametersHelper {
+
+    private CameraParametersHelper() {
+    }
+
+    static boolean isFocusModeSupported(String focusMode, Camera camera) {
```
EarlOfEgo: even though it's package private, it should use support annotations

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraParametersHelper.java
```diff
@@ -0,0 +1,27 @@
...
+        return camera.getParameters().getSupportedFocusModes().contains(focusMode);
+    }
+
+    static boolean isUsingFocusMode(String focusMode, Camera camera) {
```
EarlOfEgo: same as above, e.g. will crash if camera is null

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraParametersHelper.java
```diff
@@ -0,0 +1,27 @@
...
+        return camera.getParameters().getFocusMode().equals(focusMode);
+    }
+
+    static boolean isFlashModeSupported(String flashMode, Camera camera) {
```
EarlOfEgo: ⬆️ same

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraParametersHelper.java
a-szotyori: Thanks, really good catch! I added the annotations.

ginivision/src/main/java/net/gini/android/vision/internal/camera/api/CameraController.java
```diff
@@ -41,7 +47,8 @@
 
     private boolean mPreviewRunning = false;
     private AtomicReference<CompletableFuture<Boolean>> mFocusingFuture = new AtomicReference<>();
-    private AtomicReference<CompletableFuture<Photo>> mTakingPictureFuture = new AtomicReference<>();
```
nsobadjiev: I really wish the commit wasn't full of those line breaks. It's distracting and pushes attention away from the important changes. I'd prefer introducing line length limits to be a separate commit.

a-szotyori:
Added tests for verifying the new behaviour.

PR #32, opened 2017-01-27
Screen API: back button leads back to the previous Activity

a-szotyori:
One of our clients rightfully noticed, that our back button (either on the bottom or on the ActionBar) closes the library in the Screen API instead of leading back to the previous Activity.

We fixed this now and also added a boolean extra to the `CameraActivity` to revert to the previous behaviour, if required.

### How to test
Run the Screen API example app and test the back button functionality. Both the bottom back button and the one on the ActionBar should be tested.

To test the previous behaviour uncomment in the `MainActivity#startGiniVisionLibrary()` method the line containing the `EXTRA_IN_BACK_BUTTON_SHOULD_CLOSE_LIBRARY`.

### Merging
Nothing special.

### Ticket 
https://tickets.i.gini.net/issue/MOBILE-701

### DIFF
--- a/ginivision/src/androidTest/java/net/gini/android/vision/camera/CameraScreenTest.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/camera/CameraScreenTest.java
@@ -6,6 +6,10 @@ import static net.gini.android.vision.OncePerInstallEventStoreHelper.setOnboardi
 import static net.gini.android.vision.test.EspressoMatchers.hasComponent;
 import static net.gini.android.vision.test.Helpers.prepareLooper;
 
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.argThat;
+import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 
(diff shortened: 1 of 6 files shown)


ginivision/src/androidTest/java/net/gini/android/vision/camera/CameraScreenTest.java
```diff
@@ -238,29 +269,67 @@ public void should_passAnalysisActivityIntent_toReviewActivity() throws Interrup
...
-    @NonNull
-    private Intent getCameraActivityIntent() {
-        Intent intent = new Intent(Intent.ACTION_MAIN);
-        CameraActivity.setReviewActivityExtra(intent, InstrumentationRegistry.getTargetContext(),
-                ReviewActivityTestStub.class);
-        CameraActivity.setAnalysisActivityExtra(intent, InstrumentationRegistry.getTargetContext(),
-                AnalysisActivityTestStub.class);
-        return intent;
+    @Test
+    public void should_passBackButtonClosesLibraryExtra_toReviewActivity()
```
nsobadjiev: Isn't this method too long for a unit test? It's hard to find out what it does.

a-szotyori: True, I'll refactor it.

screenapiexample/src/main/java/net/gini/android/vision/screen/MainActivity.java
```diff
@@ -92,6 +92,10 @@ private void startGiniVisionLibrary() {
         // Set EXTRA_IN_SHOW_ONBOARDING to true, to show the OnboardingActivity when the CameraActivity starts
         //intent.putExtra(CameraActivity.EXTRA_IN_SHOW_ONBOARDING, true);
 
+        // Set EXTRA_IN_BACK_BUTTON_SHOULD_CLOSE_LIBRARY to true, to close library on pressing the back
+        // button from any Activity in the library
+        //intent.putExtra(CameraActivity.EXTRA_IN_BACK_BUTTON_SHOULD_CLOSE_LIBRARY, true);
```
EarlOfEgo: Having so many configurations it could make sense to create a builder to generate the intent. Or maybe at least static factory methods. Then the static strings can be hidden and no one has to write the above code 😛

PR #36, opened 2017-03-10
Adding rotation delta to EXIF User Comment field

a-szotyori:
In order to detect images which were uploaded and then uploaded again due to rotation we add the rotation delta to the rotated image. Rotation delta is the degrees rotated compared to the image from the camera.

This PR also contains a fix regarding generating the UUID for the User Comment: The bug was that a new UUID was generated whenever a Photo instance was created from a Document.

### How to test
Run the tests in the `ginivision/src/androidTest` folder. Specifically the following tests are relevant: 
* `net.gini.android.vision.internal.camera.photo.PhotoTest`
* `net.gini.android.vision.review.ReviewScreenTest`

### Merging
Nothing special.

### Ticket 
No ticket.

### DIFF
--- a/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
@@ -88,4 +88,72 @@ public class PhotoTest {
         // Then
         assertAbout(photo()).that(photo).hasUUIDinUserComment(uuid);
     }
+
+    @Test
+    public void should_initRotationDelta_whenCreated() {
+        // When
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // Then
+        assertThat(photo.getRotationDelta()).isEqualTo(0);
+    }
+
+    @Test
+    public void should_addRotationDelta_toExifUserComment() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(0);
+    }
+
+    @Test
+    public void should_updateRotationDelta_afterCWRotation() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotate(90).apply();
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(90);
+    }
+
+    @Test
+    public void should_updateRotationDelta_afterCCWRotation() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotate(-90).apply();
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(270);
+    }
+
+    @Test
+    public void should_cumulateRotationDelta_afterMultipleRotations() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotate(90).apply();
+        photo.edit().rotate(90).apply();
+        photo.edit().rotate(-90).apply();
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(90);
+    }
+
+    @Test
+    public void should_normalizeRotationDelta_forCWRotation() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotate(450).apply();
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(90);
+    }
+
+    @Test
+    public void should_normalizeRotationDelta_forCCWRotation() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotate(-270).apply();
+        // Then
+        assertAbout(photo()).that(photo).hasRotationDeltaInUserComment(90);
+    }
 }
--- a/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
@@ -9,6 +9,8 @@ import static net.gini.android.vision.test.PhotoSubject.photo;
 
 import android.support.test.runner.AndroidJUnit4;
 
+import net.gini.android.vision.Document;
+
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
@@ -40,6 +42,37 @@ public class PhotoTest {
         Photo photoFromParcel = doParcelingRoundTrip(photo, Photo.CREATOR);
         // Then
         assertThat(photoFromParcel).isEqualTo(photo);
+    }
+
+    @Test
+    public void should_keepUserComment_whenCreating_fromDocument() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        Photo fromDocument = Photo.fromDocument(Document.fromPhoto(photo));
+        // Then
+        assertAbout(photo()).that(photo).hasSameUserCommentAs(fromDocument);
+    }
+
+    @Test
+    public void should_setUUIDfromUserComment_whenCreating_fromDocument() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        Photo fromDocument = Photo.fromDocument(Document.fromPhoto(photo));
+        // Then
+        assertThat(photo.getUUID()).isEqualTo(fromDocument.getUUID());
+    }
+
+    @Test
+    public void should_setRotationDeltafromUserComment_whenCreating_fromDocument() {
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        photo.edit().rotateTo(90).apply();
+        Photo fromDocument = Photo.fromDocument(Document.fromPhoto(photo));
+        // Then
+        assertThat(photo.getRotationDelta()).isEqualTo(fromDocument.getRotationDelta());
     }
 
     @Test
--- a/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
@@ -55,13 +55,13 @@ public class PhotoTest {
     }
 
     @Test
-    public void should_setUUIDfromUserComment_whenCreating_fromDocument() {
+    public void should_setContentIdFromUserComment_whenCreating_fromDocument() {
         // Given
         Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
         // When
         Photo fromDocument = Photo.fromDocument(Document.fromPhoto(photo));
         // Then
-        assertThat(photo.getUUID()).isEqualTo(fromDocument.getUUID());
+        assertThat(photo.getContentId()).isEqualTo(fromDocument.getContentId());
     }
 
     @Test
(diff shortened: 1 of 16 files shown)


ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
```diff
@@ -43,6 +45,37 @@ public void should_supportParceling() {
     }
 
     @Test
+    public void should_keepUserComment_whenCreating_fromDocument() {
```
a-szotyori: This is related to the UUID fix, but is also needed for the rotation delta.

The Document class contains the image and is part of the public API. The Photo class is used only internally.

The Screen API uses the Component API so we share the image in a Document instance between the Fragments of the Component API and then convert it back to a Photo instance internally.

ginivision/src/androidTest/java/net/gini/android/vision/internal/camera/photo/PhotoTest.java
```diff
@@ -43,6 +45,37 @@ public void should_supportParceling() {
...
+        // Given
+        Photo photo = Photo.fromJpeg(TEST_JPEG, 0);
+        // When
+        Photo fromDocument = Photo.fromDocument(Document.fromPhoto(photo));
+        // Then
+        assertAbout(photo()).that(photo).hasSameUserCommentAs(fromDocument);
+    }
+
+    @Test
+    public void should_setUUIDfromUserComment_whenCreating_fromDocument() {
```
a-szotyori: This is related to the UUID bugfix, which I didn't discover in the previous PR.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
```diff
@@ -33,7 +33,15 @@
 /**
  * @exclude
  */
-public class Exif {
+class Exif {
+
+    static final String USER_COMMENT_MAKE = "Make";
```
a-szotyori: It was high time to extract the keys into constants...

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ExifReader.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.commons.imaging.formats.tiff.TiffField;
+import org.apache.commons.imaging.formats.tiff.constants.ExifTagConstants;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+/**
+ * @exclude
+ */
+class ExifReader {
```
a-szotyori: This is related to the UUID bugfix, but is also needed for the rotation delta.

We need to read the Exif UserComment to initialize the fields of a Photo instance when created from a Document.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
```diff
@@ -29,43 +29,77 @@
...
-    public static Photo fromJpeg(@NonNull byte[] jpeg, int orientation) {
-        Photo photo = new Photo();
-        photo.setJpeg(jpeg);
-        photo.setBitmapPreview(createPreview(jpeg));
-        photo.setRotationForDisplay(orientation);
-        photo.readRequiredTags();
-        photo.updateExif();
-        return photo;
+    public static Photo fromJpeg(@NonNull final byte[] jpeg, final int orientation) {
+        return new Photo(jpeg, orientation);
```
a-szotyori: Refactored this, because the steps done in this factory method are required for the construction of a Photo instance so they should go into the constructor.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
```diff
@@ -29,43 +29,77 @@
...
     }
 
-    public static Photo fromDocument(@NonNull Document document) {
-        return Photo.fromJpeg(document.getJpeg(), document.getRotationForDisplay());
+    public static Photo fromDocument(@NonNull final Document document) {
+        return new Photo(document);
```
a-szotyori: Same as above.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
```diff
@@ -81,17 +115,23 @@ public synchronized int getRotationForDisplay() {
...
-        // Converts input degrees to degrees between [0,360]
+    synchronized void setRotationForDisplay(int degrees) {
+        // Converts input degrees to degrees between [0,360)
         mRotationForDisplay = ((degrees % 360) + 360) % 360;
     }
 
-    public synchronized void setRotationDelta(int degrees) {
-        // Converts input degrees to degrees between [0,360]
-        mRotationDelta = ((degrees % 360) + 360) % 360;
+    synchronized void updateRotationDeltaBy(int degrees) {
```
a-szotyori: Renamed, to make the intent clear, that the rotation delta change is cumulative.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoCompressionModifier.java
```diff
@@ -0,0 +1,39 @@
...
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.support.annotation.NonNull;
+
+import java.io.ByteArrayOutputStream;
+
+/**
+ * @exclude
+ */
+class PhotoCompressionModifier implements PhotoModifier {
```
a-szotyori: Refactored PhotoEdit to be extensible less error prone, by abstracting away the modifications into classes.

This one does the compression.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoEdit.java
```diff
@@ -1,88 +1,70 @@
...
     private final Photo mPhoto;
-    private int mRotationDegrees;
-    private int mQuality = DEF_QUALITY;
+    private final List<PhotoModifier> mPhotoModifiers;
 
-    public PhotoEdit(@NonNull Photo photo) {
+    PhotoEdit(@NonNull Photo photo) {
         mPhoto = photo;
-        resetToDefaults();
+        mPhotoModifiers = new ArrayList<>();
```
a-szotyori: As mentioned above modifications were abstracted away. For each modification a new instance of a specific modifier is created and added to this list. When the modification is applied the list is iterated and the modifications are applied in order.

I kept it really simple.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoEdit.java
```diff
@@ -1,88 +1,70 @@
...
     }
 
     @NonNull
-    public PhotoEdit rotate(int degrees) {
-        degrees %= 360;
-        mRotationDegrees = degrees;
+    public PhotoEdit rotateTo(int degrees) {
```
a-szotyori: Clarified the intent to show that the image is rotated to the given degrees. So if it was at 90 and you call rotateTo(180) the image won't be orientated to 270 degrees but 180.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoEdit.java
```diff
@@ -1,88 +1,70 @@
...
+        final PhotoRotationModifier rotationModifier = new PhotoRotationModifier(degrees, mPhoto);
+        mPhotoModifiers.add(rotationModifier);
         return this;
     }
 
     @NonNull
-    public PhotoEdit compress(int quality) {
-        mQuality = quality;
+    public PhotoEdit compressBy(int quality) {
```
a-szotyori: Clarified the intent to show that the compression is applied to the any existing compression.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoRotationModifier.java
```diff
@@ -0,0 +1,29 @@
+package net.gini.android.vision.internal.camera.photo;
+
+import android.support.annotation.NonNull;
+
+/**
+ * @exclude
+ */
+class PhotoRotationModifier implements PhotoModifier {
```
a-szotyori: This modifier only updates the exif orientation and the rotation delta.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/PhotoModifier.java
```diff
@@ -0,0 +1,8 @@
+package net.gini.android.vision.internal.camera.photo;
+
+/**
+ * @exclude
+ */
+interface PhotoModifier {
```
a-szotyori: Common interface for any photo modification implementations.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ExifReader.java
```diff
@@ -0,0 +1,63 @@
...
+            throw new ExifReaderException("Could not read jpeg metadata: " + e.getMessage(), e);
+        }
+    }
+
+    @Nullable
+    String getValueForKeyfromUserComment(@NonNull final String key, @NonNull final String userComment) {
+        final String[] keyValuePairs = userComment.split(",");
+        for (final String keyValuePair : keyValuePairs) {
+            final String[] keyAndValue = keyValuePair.split("=");
+            if (keyAndValue[0].equals(key)) {
```
nsobadjiev: Is referencing the 0 and 1 index directly safe here? I know we usually enforce a specific format for the user comment, but it's still an external EXIF field. And if users of out library choose to feed us custom images or add something themselves, an exception would be thrown.

ginivision/src/androidTest/java/net/gini/android/vision/review/ReviewScreenTest.java
```diff
@@ -416,4 +415,113 @@ public void onProceedToAnalysisScreen(@NonNull final Document document) {
...
+        });
+
+        // Allow the activity to run a little for listeners to be invoked
+        Thread.sleep(PAUSE_DURATION);
+
+        assertAbout(document()).that(documentToAnalyze.get()).hasRotationDeltaInUserComment(0);
+    }
+
+    @Test
+    public void should_returnDocument_withNonZeroRotationDelta_inProceedToAnalysis_ifDocumentWasRotated()
```
a-szotyori: Yeah, debugging might be an issue. These are white box Activity integration tests (or white box Activity UI tests) to verify that the library's Activities fulfil our public API contract.

Back when we created the library we implemented these tests to verify that the intended behaviour is fulfilled.

You are right, ideally we would need to cover the behaviour of the library's public API in unit tests and these high-level tests should serve the role of API contract acceptance testing.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
```diff
@@ -364,30 +367,34 @@ private String createUserComment() {
...
             }
             // Model
             if (mAddModel) {
-                userCommentBuilder.append("Model=");
+                userCommentBuilder.append(USER_COMMENT_MODEL).append("=");
                 userCommentBuilder.append(Build.MODEL);
                 userCommentBuilder.append(",");
             }
             // Platform
-            userCommentBuilder.append("Platform=Android");
```
a-szotyori: 👍 Will do.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ExifReader.java
a-szotyori: Good catch! Fortunately the split method never returns an empty list. It returns the whole string if couldn't be split.

Still I added a length check and a test.

a-szotyori:
Implemented some review remarks and renamed `UUID` to `ContentId`.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ExifReader.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    @Nullable
+    String getValueForKeyFromUserComment(@NonNull final String key, @NonNull final String userComment) {
+        final String[] keyValuePairs = userComment.split(",");
+        for (final String keyValuePair : keyValuePairs) {
+            final String[] keyAndValue = keyValuePair.split("=");
+            if (keyAndValue.length > 0 && keyAndValue[0].equals(key)) {
+                return keyAndValue[1];
```
nsobadjiev: But you still reference `keyAndValue[1]`. So if the length is 1, that would be a problem. As you said, if split cannot find the separator, it would return one element.

a-szotyori: 🙈

PR #37, opened 2017-06-28
Bugfix exif npe missing user comment

a-szotyori:
Fixed ExifReader bugs:
* Fixed NPE when user comment was missing.
* Fixed array copy when user comment was too short.

Wrote additional tests for the ExifReader.

### How to test
Run the tests in `ginivision/src/androidTest`: 
`$ ./gradlew ginivision:connectedAndroidTest`


ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ExifReader.java
```diff
@@ -25,7 +25,7 @@
     }
 
     @NonNull
-    public String getUserComment() {
+    String getUserComment() {
         try {
             final JpegImageMetadata jpegMetadata = (JpegImageMetadata) getMetadata(jpeg);
```
ricogini: Why holds the ExifReader the jpeg byte array? Would it be better to have a separate ExifReader factory method, which extracts the JpegImageMetadata and creates a new ExifReader from it which could read the separate fields?

PR #39, opened 2017-07-04
Guide for open with

a-szotyori:
A guide for adding "Open with..." for PDFs and images to an Android app.

### DIFF
--- a/ginivision/src/doc/source/guide-for-open-with.rst
+++ b/ginivision/src/doc/source/guide-for-open-with.rst
@@ -45,7 +45,7 @@ Since the introduction of `Run Time Permissions <https://developer.android.com/t
  
 We recommend checking for the permission whenever your app receives a PDF or an image and before accessing it. If an explanation is requested by the system to be shown to the user we highly recommend providing one.
  
-Handling the case where permission is not granted is required.  We recommend showing an explanation and providing a link to your app’s *App info* page in the *Settings* app with a short guide on how to grant the permission there. 
+It is required to handle the case where the user did not grant permissions. We recommend showing an explanation and providing a link to your app’s *App info* page in the *Settings* app with a short guide on how to grant the permission there. 
  
 The Intent for opening the *App info* page can be created as follows:
 
@@ -114,7 +114,7 @@ Getting the MIME type from the Intent:
         }
     }
 
-Having the ``Uri`` the contents of the PDF or image files can be read using the ``ContentResolver``.
+Having the ``Uri``, the contents of the PDF or image files can be read using the ``ContentResolver``.
  
 Reading the PDF or image into a byte array:
 
--- a/ginivision/src/doc/source/guide-for-open-with.rst
+++ b/ginivision/src/doc/source/guide-for-open-with.rst
@@ -1,0 +1,182 @@
+How to Upload PDFs and Images
+=============================
+
+General considerations
+----------------------
+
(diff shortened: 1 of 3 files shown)


ginivision/src/doc/source/guide-for-open-with.rst
```diff
@@ -0,0 +1,182 @@
+How to Upload PDFs and Images
+=============================
+
+General considerations
+----------------------
+
+Enabling your app to open PDFs and images allows your users to open any kind of files which are identified by the OS as PDFs or images. Make sure to check the MIME type and that the file size is below an acceptable threshold (5MB for example). It is also advisable to check the first bytes of the incoming files to determine their type and allow only PDFs and known image types.
```
a-benson: Please see comment in iOS version: if we only want to allow single-page documents and a check for this is possible, please add it as a suggested check under General considerations.

ginivision/src/doc/source/guide-for-open-with.rst
```diff
@@ -0,0 +1,182 @@
...
+
+.. code-block:: xml
+
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+ 
+Since the introduction of `Run Time Permissions <https://developer.android.com/training/permissions/requesting.html>`_ with Android 6.0 (API Level 23) your app also has to request permission from the user before accessing a PDF or an image file outside of the app-specific directories. 
+ 
+We recommend checking for the permission whenever your app receives a PDF or an image and before accessing it. If an explanation is requested by the system to be shown to the user we highly recommend providing one.
+ 
+Handling the case where permission is not granted is required.  We recommend showing an explanation and providing a link to your app’s *App info* page in the *Settings* app with a short guide on how to grant the permission there.
```
a-benson: Suggested wording for first sentence: "It is required to handle the case where the user did not grant permissions."

ginivision/src/doc/source/guide-for-open-with.rst
```diff
@@ -0,0 +1,182 @@
...
+        if (clipData != null) {
+            ClipDescription description = clipData.getDescription();
+            for (int i = 0; i < description.getMimeTypeCount(); i++) {
+                type = description.getMimeType(i);
+                mimeTypes.add(type);
+            }
+        }
+    }
+
+Having the ``Uri`` the contents of the PDF or image files can be read using the ``ContentResolver``.
```
a-benson: Suggest adding a comma after "Uri". So: "Having the 'Uri', the contents of the PDF or image files can be read using the 'ContentResolver'."

ginivision/src/doc/source/guide-for-open-with.rst
```diff
@@ -0,0 +1,182 @@
...
+The byte array from the example above can be directly uploaded to the Gini API for information extraction.
+
+Documentation
+^^^^^^^^^^^^^
+
+- Android: https://developer.android.com/training/sharing/receive.html
+- Gini API: http://developer.gini.net/gini-api/html/documents.html#submitting-files
+- Gini API SDK: http://developer.gini.net/gini-sdk-android/guides/common-tasks.html#upload-a-document
+
+Showing a preview of the PDF’s first page
```
a-benson: Please see comment from iOS documentation version. Since we don't currently support multipage scan, do we want to only speak of a preview for the PDF/image, instead of the first page of the PDF? And also making this consistent in the following text.

Issue #55, opened 2017-08-24
Add device orientation and type to User Comment Exif field


a-szotyori:
Key name in the User Comment csv should be `DeviceOrientation` and the value should be `portrait` or `landscape`. 

For device type the key name should be `DeviceType` and the value should be `phone` or `tablet`.

PR #60, opened 2017-08-25
Added device orientation and device type to the exif user comment

a-szotyori:
Device orientation landscape or portrait and device type tablet or phone are added to the user comment exif field.

### How to test
Run the screen or component api example apps and check the generated user comment string.

### Merging
Automatic.

### Ticket 
Issue #55

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
@@ -31,20 +31,26 @@ public class Photo implements Parcelable {
     private int mRotationForDisplay = 0;
     private String mContentId = "";
     private int mRotationDelta = 0;
+    private String mDeviceOrientation;
+    private String mDeviceType;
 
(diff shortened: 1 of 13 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Photo.java
```diff
@@ -301,6 +329,8 @@ public int hashCode() {
         result = 31 * result + mRotationForDisplay;
         result = 31 * result + (mContentId != null ? mContentId.hashCode() : 0);
         result = 31 * result + mRotationDelta;
+        result = 31 * result + (mDeviceOrientation != null ? mDeviceOrientation.hashCode() : 0);
+        result = 31 * result + (mDeviceType != null ? mDeviceType.hashCode() : 0);
         return result;
```
kikettas: Why is it multiplied by 31?

a-szotyori: Could be any number, I let Android Studio generate these.

Issue #56, opened 2017-08-24
Issues getting extractions in portrait orientation on Lenovo Tab 2 10.1"


a-szotyori:
During our testing session we failed to get any correct extractions on the Lenovo Tab 2 10.1" in portrait orientation.

None of the other tablets manifested this issue.

PR #61, opened 2017-08-27
Bugfix lenovo tab 2 portrait

a-szotyori:
On the Lenovo Tab 2 10.1" we got extractions only in landscape. The cause for this was that the exif tags model and make contained null bytes and in this case the Commons Imaging library returned a string array instead of a string which caused the exif manipulation to fail. Due to this the correct orientation was not added to the exif and the Gini API could not rotate the image to the correct orientation.

### How to test
Run the tests in ExifTest and try the example apps on the Lenovo Tab 2 10.1".

### Merging
Automatic.

### Ticket 
Issue #56


ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
```diff
@@ -224,10 +224,46 @@ public Exif build() {
 
         private void addStringExif(@NonNull TiffOutputDirectory outputDirectory,
                 @NonNull TiffField field) throws ImageReadException {
-            byte bytes[] = field.getStringValue().getBytes(Charset.forName("US-ASCII"));
+            byte bytes[] = getStringFieldValue(field).getBytes(Charset.forName("US-ASCII"));
             addStringExif(outputDirectory, field.getTagInfo(), bytes);
         }
 
+        @NonNull
+        private String getStringFieldValue(@NonNull final TiffField field) throws ImageReadException {
```
kikettas: Could it be replaced with a toString or similar method on TiffField class?

a-szotyori: Unfortunately not since we don't initialize the instance, we receive it from `JpegImageMetadata#findEXIFValue()`.

By analysing the implementation again I discovered, that handling make and model tags as strings is not needed, we can simply add them like the other ones and eliminate a potential source of bugs completely.

Issue #65, opened 2017-09-07
No Results Screen - Screen API


a-szotyori:
Requirements:
- Screen API: 
  - an Activity hosting the Component API's Fragment
  - client notifies the Analysis Screen that no extractions were received and GVL shows the No Results Screen automatically
- User interaction:
  - User can ~~close GVL~~ go back to the Camera Screen with either the top left "up button" or the back button.

a-szotyori:
Up and back buttons lead back to the Camera Screen instead of closing GVL.

Issue #64, opened 2017-09-07
No Results Screen - Component API


a-szotyori:
Requirements:
- Component API: 
  - a Fragment containing the No Results Screen
  - client shows the No Results Fragment manually and gets notified through the fragment's listener about user interaction
- User interaction: 
  - User can go back to the Camera Screen with a button on the screen
  - Back button is not handled by the fragment
- UI:
  - support phone and tablet layouts
  - phones: limit to portrait orientation
  - tablets: allow landscape orientation
  - mockup (tablet design not yet available):
![gini_pay_android_000_camera_roll_008_rechnung_richtig_fotografieren](https://user-images.githubusercontent.com/13149498/30166771-9f03ec1e-93e4-11e7-9fee-ed1323d3386d.png)

PR #70, opened 2017-09-08
Feature/no results screens

EarlOfEgo:
Implemented the no results screens, created Fragments for compat and normal version, see #64 for more information. As well as layout files for phone, tablet land- and portrait versions. However the used assets, dimensons, colors and everything are just mockups since the final stuff arrives.

Added also an activity which hosts the fragments in the screen api, see #65 
### How to test
Don't know yet. Maybe use screen example and take a picture of something that doesn't have extractions.

### Merging
Since everything is a mockup, don't know if it makes sense to merge now.


ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentInterface.java
```diff
@@ -11,17 +11,20 @@
...
      * </p>
      */
-    void startScanAnimation();
+    void hideError();
 
     /**
      * <p>
-     *     You should call this method when the document analysis has finished.
+     *     You should call this method after you've received the analysis results from the Gini
+     *     API and there have been no extractions found.
```
a-szotyori: What do you think about `You should call this method after you've received the analysis results from the Gini API without the required extractions.` ?

ginivision/src/main/java/net/gini/android/vision/noresults/NoResultsFragmentCompat.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Nullable
+    @Override
+    public View onCreateView(final LayoutInflater inflater, @Nullable final ViewGroup container,
+            final Bundle savedInstanceState) {
+        return mFragmentImpl.onCreateView(inflater, container, savedInstanceState);
+    }
+
+    public static NoResultsFragmentCompat newInstance() {
```
a-szotyori: Should be `createInstance()`, we used `createInstance...` for the other fragments.

ginivision/src/main/java/net/gini/android/vision/noresults/NoResultsFragmentStandard.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Nullable
+    @Override
+    public View onCreateView(final LayoutInflater inflater, @Nullable final ViewGroup container,
+            final Bundle savedInstanceState) {
+        return mFragmentImpl.onCreateView(inflater, container, savedInstanceState);
+    }
+
+    public static NoResultsFragmentStandard newInstance() {
```
a-szotyori: Same as before `createInstance`

screenapiexample/src/main/java/net/gini/android/vision/screen/AnalysisActivity.java
```diff
@@ -61,17 +57,26 @@ public void onClick(View v) {
                     }
                 });
             }
+
+            @Override
+            public void onExtractionsReceived(Map<String, SpecificExtraction> extractions) {
+                mExtractions = extractions;
+                if (mExtractions == null || mExtractions.isEmpty()) {
```
a-szotyori: I think a lot of the time we get some extractions, like `invoiceType` or similar even when there are no pay 5 extractions. You could use a method similar to `ExtractionsActivity#isPay5Extraction()` to check whether we got at least one pay 5 extraction.

screenapiexample/src/main/java/net/gini/android/vision/screen/AnalysisActivity.java
```diff
@@ -61,17 +57,26 @@ public void onClick(View v) {
...
+                    noExtractionsFound();
```
a-szotyori: We should also add this to the Component API example app.

a-szotyori:
As we discussed up and back should lead back to the camera screen.

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisActivity.java
```diff
@@ -159,6 +160,8 @@ public void hideError() {
     public void noExtractionsFound() {
         final Intent noResultsActivity = new Intent(this, NoResultsActivity.class);
         startActivity(noResultsActivity);
+        setResult(Activity.RESULT_CANCELED);
```
a-szotyori: I think it's clearer if we set an own result code here, like `RESULT_NO_EXTRACTIONS`.

ginivision/src/main/java/net/gini/android/vision/review/ReviewActivity.java
```diff
@@ -321,6 +321,8 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                 setResult(resultCode, data);
                 finish();
                 clearMemory();
+            } else {
```
a-szotyori: And then we check for that result code here and it would be clearer what our intent is I think:
```
            if (resultCode == AnalysisActivity.RESULT_NO_EXTRACTIONS) {
                finish();
                clearMemory();
            } else if (mBackButtonShouldCloseLibrary
                    || resultCode != Activity.RESULT_CANCELED) {
                setResult(resultCode, data);
                finish();
                clearMemory();
            }
```

ginivision/src/main/java/net/gini/android/vision/review/ReviewActivity.java
```diff
@@ -321,6 +321,8 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                 setResult(resultCode, data);
```
a-szotyori: I know that `mBackButtonShouldCloseLibrary` confuses things. We used that when we changed the behavior to allow clients to revert to the old behavior. I'll create an issue to flag it deprecated and remove it in a few months.

ginivision/src/main/java/net/gini/android/vision/review/ReviewActivity.java
EarlOfEgo: I see that `RESULT_NO_EXTRACTIONS` can makes sense [here](https://github.com/gini/gini-vision-lib-android/pull/70#discussion_r138323423) but then we have to override the back/up button behaviour as well. Those both set `RESULT_CANCELED` when used.

a-szotyori: They don't matter in this case, since on no extractions, the analysis screen is finished with `RESULT_NO_EXTRACTIONS`. If back/up is used, nothing should change.

Issue #68, opened 2017-09-07
Show and cycle through suggestions after a delay


a-szotyori:
If the Analysis Screen is visible for 5s show the first randomly picked suggestion.

Cycle through suggestions with 4s display time for each one.

Transitions should be animated.

PR #81, opened 2017-09-15
Feature show cycle suggestions

EarlOfEgo:
Added animation for showing the hints in a random order. They are sliding in from the bottom and slide out there.
The assets aren't there yet and have to be replaced later, same with the sizes and other things which are UI related.

### How to test
upload a document that takes pretty long to process and look at some randomly ordered hints which are animated sliding in.

### Merging
none

### Ticket 
#68

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
+++ b/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
@@ -39,8 +39,6 @@ import java.util.Random;
 
 class AnalysisFragmentImpl implements AnalysisFragmentInterface {
 
-    public static final int HINT_START_DELAY = 5000;
-    public static final int HINT_CYCLE_INTERVAL = 4000;
     private static final AnalysisFragmentListener NO_OP_LISTENER = new AnalysisFragmentListener() {
         @Override
         public void onAnalyzeDocument(@NonNull Document document) {
@@ -66,6 +64,8 @@ class AnalysisFragmentImpl implements AnalysisFragmentInterface {
     private Runnable mRunnable;
 
     private static final int HINT_ANIMATION_DURATION = 500;
+    private static final int HINT_START_DELAY = 5000;
+    private static final int HINT_CYCLE_INTERVAL = 4000;
 
 
     public AnalysisFragmentImpl(FragmentImplCallback fragment, Document document, String documentAnalysisErrorMessage) {
--- a/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
+++ b/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
@@ -2,7 +2,9 @@ package net.gini.android.vision.analysis;
 
 import static net.gini.android.vision.internal.util.ActivityHelper.forcePortraitOrientationOnPhones;
 
+import android.app.Activity;
 import android.content.Context;
+import android.graphics.Point;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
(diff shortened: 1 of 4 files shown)


ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -110,15 +133,111 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+                    public void onAnimationCancel(final View view) {
+                    }
+                });
+    }
+
+    private int getContainerHeight() {
+        final Activity activity = mFragment.getActivity();
+        final Display display = activity.getWindowManager().getDefaultDisplay();
+        final Point size = new Point();
+        display.getSize(size);
```
a-szotyori: If the fragment is not shown full screen, we might not get the size we are looking for. What about getting the layout root and getting it's height? Would have to use the ViewTreeObserver to get the final height if there's no better way you know of...

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -110,15 +133,111 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+                .translationY(0)
+                .setDuration(HINT_ANIMATION_DURATION)
+                .setListener(new ViewPropertyAnimatorListener() {
+                    @Override
+                    public void onAnimationStart(final View view) {
+                    }
+
+                    @Override
+                    public void onAnimationEnd(final View view) {
+                        mHandler.postDelayed(mRunnable, HINT_CYCLE_INTERVAL);
```
a-szotyori: `mRunnable` is too generic. The var name should tell us what this runnable does.

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -110,15 +133,111 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+        final Context context = mFragment.getActivity();
+        mHintImageView.setImageDrawable(ContextCompat.getDrawable(context, nextHint.getDrawableResource()));
+        mHintTextView.setText(nextHint.getTextResource());
+    }
+
+    private AnalysisHint getNextHint() {
+        if(mHints.isEmpty()) {
+            mHints = generateRandomHintsList();
+        }
+        return mHints.remove(0);
```
a-szotyori: By reshuffling after all hints were shown we could end up showing the same hint twice one after the other. Wouldn't it be enough to shuffle it once and then loop through the list?

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisHint.java
```diff
@@ -0,0 +1,42 @@
...
+import android.support.annotation.DrawableRes;
+import android.support.annotation.StringRes;
+
+import net.gini.android.vision.R;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+enum AnalysisHint {
```
a-szotyori: 👍 Me like!

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -110,15 +133,111 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+                mHintAnimation = getSlideDownAnimation(containerHeight);
+                mHintAnimation.start();
+            }
+        };
+        mHandler.postDelayed(mRunnable, HINT_START_DELAY);
+    }
+
+    @NonNull
+    private ViewPropertyAnimatorCompat getSlideDownAnimation(final int containerHeight) {
+        return ViewCompat.animate(mHintContainer)
```
a-szotyori: You could try using the support TransitionManager. We wouldn't need to get the container height.

Issue #63, opened 2017-09-01
File type check


a-szotyori:
We may only accept files fulfilling the following conditions:

- mime type `image/jpeg`, `image/png`, `image/tiff`, `image/gif` and `application/pdf` http://developer.gini.net/gini-api/html/index.html
- size less than 10MB
- if PDF than 10 pages or less
- (optional) magic bytes checks https://blog.netspi.com/magic-bytes-identifying-common-file-formats-at-a-glance/

PR #82, opened 2017-09-19
Implemented criteria checking for the imported files.

EarlOfEgo:
Check if the chosen file is supported and matches the criteria.
The problem right now is that without a third party library we can't check if a PDF has <= 10 pages for SDK versions < 21. However to add another library would be wrong and implementing by our own would be an overkill in my opinion, therefore this check always succeeds now.

### How to test
open some files and see the toast

### Merging
none

### Ticket 
#63

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/fileimport/FileChooserActivity.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/fileimport/FileChooserActivity.java
@@ -9,8 +9,10 @@ import static net.gini.android.vision.GiniVisionError.ErrorCode.DOCUMENT_IMPORT;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ResolveInfo;
+import android.graphics.pdf.PdfRenderer;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.ParcelFileDescriptor;
 import android.provider.MediaStore;
 import android.support.annotation.NonNull;
 import android.support.transition.AutoTransition;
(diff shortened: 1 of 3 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/fileimport/FileChooserActivity.java
```diff
@@ -144,6 +155,57 @@ protected void onActivityResult(final int requestCode, final int resultCode,
         finish();
     }
 
+    private boolean fileMatchesCriteria(final Intent data) {
```
a-szotyori: I would be happier if we would have an extensible and independent file checking solution. For "Open with" we would want to check the file in the Review Screen or the Analysis Screen. We might also add a new criteria or remove an existing one. For showing relevant error information to users we also would need what was wrong: too big, too many pages, ...

I drew up a little sketch of how that could look like:
file-validation

`FileValidator` is used as the entry point. It checks each concrete validation in turn and could return the first error. Using the `Result` callers can check for success and use the `ValidationType` to handle each particular validation failure as required.

What do you think?

EarlOfEgo: I think that might be a little bit over engineered. Sure there can be a util class that can do the file checks, and returning true or false(and/or throwing an appropriate error).
The SDK is already pretty big for what it does and adding so much logic and files just to check some simple rules might be using a sledgehammer to crack a nut.

a-szotyori: It doesn't have to be like I drew it. Something smaller is also ok. It just should be stand alone so it can be used elsewhere, too. I wouldn't glue it to the file chooser. For the "Open with" we also need to do this file check.

PR #90, opened 2017-09-22
Feature tip popover

EarlOfEgo:
Added the pop over that should hint the user to the upload functionality. The popover is shown once and then never again, therefore simple shared preferences have been used.

### How to test
open the camera screen and check if the view is there, close the view. Then kill the app and open it again, nothing should be shown.

### Merging
~do not merge yet since the close asset is missing so far.~ none

### Ticket 
#89


ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -50,6 +54,8 @@
 
 class CameraFragmentImpl implements CameraFragmentInterface {
 
+    public static final String GV_SHARED_PREFS = "GV_SHARED_PREFS";
```
a-szotyori: You can use the `OncePerInstallEventStore` with a `OncePerInstallEvent`.

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -148,6 +159,28 @@ public Object apply(Void aVoid, Throwable throwable) {
                 });
     }
 
+    private void showUploadHintPopUpOnFirstExecution() {
+        if(shouldShowHintPopUp()) {
+            ViewCompat.animate(mUploadHintContainer)
+                    .alpha(1)
+                    .setDuration(DEFAULT_ANIMATION_DURATION)
+                    .start();
+            ViewCompat.animate(mUploadHintContainer2)
```
a-szotyori: `mUploadHintContainerArrow` maybe? :)

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -324,6 +360,45 @@ public void onClick(final View view) {
...
         });
+        mUploadHintCloseButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(final View view) {
+                closeUploadHintPopUp();
+            }
+        });
+    }
+
+    private void closeUploadHintPopUp() {
```
a-szotyori: Awesome idea of using a square view rotated by 45° for the arrow! There is only one problem: the fade in/out animation reveals the covered part of the arrow:
tip-close-animation

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
EarlOfEgo: Thought of it as well, but then `OncePerInstallEventStore` & `OncePerInstallEvent` has to be public. If this is fine for you I can change it.

a-szotyori: Yeah it can be public, we flag non-public API public classes (Java packages 🤢) with @exclude in the Javadoc and won't be shown in our Javadocs...

a-szotyori:
I changed the arrow to be a shape drawable. What do you think? I like nice animations and using a drawable allows us to animate the arrow, too. But I am also fine to revert this change in case the view is safer to use on all API Levels from 14.

EarlOfEgo:
the reason I removed the animation was that it is impossible to animate two views at exactly the same time.
However if you think it looks good then use the drawables.

a-szotyori:
Thanks Stephan, we could use an AnimatorSet, but I don't think it's worth the hassle.

Issue #83, opened 2017-09-20
Update Camera Screen UI


a-szotyori:
Integrate the new screen design:
- phone: https://projects.invisionapp.com/d/main#/projects/prototypes/12011059
- tablet: https://projects.invisionapp.com/d/main#/projects/prototypes/12036352

PR #91, opened 2017-09-25
Feature update camera screen ui

EarlOfEgo:
Updated the camera screen ui. Added a shadowy frame around the focus lines, both are drawn programatically. The area inside those lines is always DIN-A4.
The drawables that used to show the focus lines have been removed and will be ignored if added by a client. With this solution it is pretty easy to extend it with the possibility to colour the focus lines differently, if wished.

### How to test
open the camera screen and see shadows

### Merging
~~do not merge yet, since the icons for uploading are black and will be replaced by white ones as soon as we get them.~~

### Ticket 
#83


ginivision/src/main/java/net/gini/android/vision/camera/FrameView.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    protected void onMeasure(final int widthMeasureSpec, final int heightMeasureSpec) {
+        mWidth = MeasureSpec.getSize(widthMeasureSpec);
+        mHeight = MeasureSpec.getSize(heightMeasureSpec);
+
+        mPortrait = mWidth < mHeight;
+
+        mWallOffsetTop = dpToPx(20);
```
a-szotyori: In landscape the the corners should be shown also in landscape. I switched setting the offsets around for landscape and it worked:
```
        if (mPortrait) {
            mWallOffsetTop = dpToPx(20);
            int documentHeight = mHeight - mWallOffsetTop;
            int documentWidth = (int) (documentHeight / Math.sqrt(2)); //din a4
            mWallOffsetSide = (mWidth - documentWidth) / 2;
        } else {
            mWallOffsetSide = dpToPx(20);
            int documentWidth = mWidth - mWallOffsetSide;
            int documentHeight = (int) (documentWidth / Math.sqrt(2)); //din a4
            mWallOffsetTop = (mHeight - documentHeight) / 2;
        }
```

PR #92, opened 2017-09-25
Feature handle imported file

a-szotyori:
Imported images and pdfs are handled by the library.

A lot of files changed because adding support for new file types required changes to some widely used classes (Document, Photo). Also new files were added for pdf handling and for managing the different file types. Runtime permission handling was also required.

**Note** Documentation was not updated. Should be done after we finalize the implementation.

Supported image types are jpeg, gif and png. Tiff is not supported by Android (for ex. BitmapFactory). Exif manipulation done only for jpegs.

All images are shown in the Review Screen, but actual orientation change is applied to jpegs only. Gifs and pngs are only shown rotated.

Pdfs are passed directly to the Analysis Screen.

File contents are loaded async in the Review Screen and Analysis Screen, whichever comes first. Exif manipulation and preview bitmap creation or pdf rendering is also done async now.

Read storage permission is requested before showing the file chooser.

### How to test
Run the app and import images and pdfs.

### Merging
Manual, lots of changes and should merge develop first, check functionality and then merge into develop.

### Ticket 
Issue #62, #80

PR #94, opened 2017-09-26
Added Image source and import method to the user comment in the exif

a-szotyori:
Added `Source` and `ImportMethod` to the EXIF User Comment. See #71 for details.

Currently comparing to the parent branch, but need to change it to `develop` after #92 was merged.

### How to test
Run the tests.

### Merging
After #92 is merged.

### Ticket 
Issue #71


ginivision/src/main/java/net/gini/android/vision/document/DocumentFactory.java
```diff
@@ -27,15 +27,17 @@
     public static Document newDocumentFromIntent(@NonNull final Intent intent,
             @NonNull final Context context,
             @NonNull final String deviceOrientation,
-            @NonNull final String deviceType) {
+            @NonNull final String deviceType,
+            @NonNull final String importMethod) {
```
EarlOfEgo: see above

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -428,7 +428,8 @@ boolean onActivityResult(final int requestCode, final int resultCode, final Inte
                     }
                     Document document = DocumentFactory.newDocumentFromIntent(data, activity,
                             DeviceHelper.getDeviceOrientation(activity),
-                            DeviceHelper.getDeviceType(activity));
+                            DeviceHelper.getDeviceType(activity),
+                            "picker");
```
EarlOfEgo: maybe an enum could make sense here, or some other constants, that would make it easier to edit/add later.

a-szotyori: I agree, I created #95 for this to avoid wider changes for now.

Issue #96, opened 2017-09-28
Show PDF filename and nr of pages in Analysis Screen


a-szotyori:
UI: https://projects.invisionapp.com/d/main#/console/12011059/253899912/inspect

PR #97, opened 2017-09-28
Show pdf filename and nr of pages in Analysis Screen

a-szotyori:
For pdf documents we show the pdf filename and nr of pages on the Analysis Screen.

### How to test
Import a pdf and check the Analysis Screen.

### Merging
Automatic, if possible.

### Ticket 
Issue #96


ginivision/src/main/java/net/gini/android/vision/internal/pdf/Pdf.java
```diff
@@ -35,6 +41,29 @@ public void toBitmap(@NonNull Size targetSize, @NonNull final Context context,
...
+            PdfRenderer pdfRenderer = null;
+            try {
+                pdfRenderer = new PdfRenderer(fileDescriptor);
+            } catch (IOException e) {
+                return 0;
+            }
+            return pdfRenderer.getPageCount();
+        }
+        return 0;
+    }
```
EarlOfEgo: I have the feeling that this is already somewhere in the code.

ginivision/src/main/java/net/gini/android/vision/internal/util/UriHelper.java
```diff
@@ -45,4 +47,28 @@
...
+     *                               available
+     */
+    @NonNull
+    public static String getFilenameFromUri(@NonNull final Uri uri, @NonNull final Context context) {
+        Cursor cursor = null;
+        cursor = context.getContentResolver().query(uri, null, null, null, null);
+        if (cursor == null) {
+            throw new IllegalStateException("Could not retrieve a Cursor for the Uri");
+        }
+        final int nameIndex = cursor.getColumnIndexOrThrow(OpenableColumns.DISPLAY_NAME);
```
EarlOfEgo: I guess this column always exists

a-szotyori: It doesn't hurt to be on the safe side with a library :)

EarlOfEgo: documentation of `getColumnIndexOrThrow` says:
>Returns the zero-based index for the given column name, or throws IllegalArgumentException if the column doesn't exist. If you're not sure if a column will exist or not use getColumnIndex(String) and check for -1, which is more efficient than catching the exceptions.

So to be on the save site maybe check with `getColumnIndex`

a-szotyori: Android... :)

a-szotyori:
Applied review suggestions.

PR #99, opened 2017-09-28
File type check error dialog

a-szotyori:
We show an error dialog in case the imported file fails validation.

**Note:** Texts are only placeholders

### How to test
Try to import an invalid file.

### Merging
Automatic.

### Ticket 
Issue #98


ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -479,31 +480,43 @@ private void requestStoragePermission(@NonNull final PermissionRequestListener l
...
+                                    activity,
+                                    DeviceHelper.getDeviceOrientation(activity),
+                                    DeviceHelper.getDeviceType(activity),
+                                    "picker");
+                            LOG.info("Document imported: {}", document);
+                            mListener.onDocumentAvailable(document);
+                        } catch (IllegalArgumentException e) {
+                            handleError(DOCUMENT_IMPORT,
+                                    "Failed to import selected document", e);
+                        }
```
EarlOfEgo: would like to see this in an extra method #cleancode

a-szotyori:
Applied review suggestions.

Issue #88, opened 2017-09-20
"Open with": Analyse documents passed to the client app


a-szotyori:
Start after #62. 

Clients can add intent filters to their apps to allow other adds to send files to their app. GVL should extend its API to allow clients to pass the external document (via intent?) to GVL to be analysed.

The flow would skip the Camera Screen. Not all documents are reviewable (like pdfs) so we should allow launching either the Review Screen or Analysis Screen directly. Here's a sketch of the planned flow: ![gvl-open-with-flow](https://user-images.githubusercontent.com/13149498/30644286-92ae3dd0-9e1a-11e7-9e6c-e243a05e3a2d.png)

PR #101, opened 2017-09-29
Feature open with

a-szotyori:
Created methods for client apps to create an Intent for the Screen API or a Document for the Component API from an "open with" Intent.

### How to test
Run the example apps and send images and pdfs from emails, photos, downloads, etc. to the example apps.

### Merging
Automatic.

### Ticket 
Issue #88

### DIFF
--- a/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
+++ b/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
@@ -18,6 +18,7 @@ import android.widget.Toast;
 
 import net.gini.android.models.SpecificExtraction;
 import net.gini.android.vision.Document;
+import net.gini.android.vision.DocumentImportEnabledFileTypes;
 import net.gini.android.vision.GiniVision;
 import net.gini.android.vision.GiniVisionCoordinator;
 import net.gini.android.vision.GiniVisionDebug;
@@ -244,7 +245,7 @@ public class GiniVisionActivity extends Activity
         if (document.isReviewable()) {
             showFragment(getReviewFragment(document), R.string.title_review);
         } else {
-            showFragment(getAnalysisFragment(document), R.string.title_analysis);
+            showFragment(getAnalysisFragment(document));
         }
     }
 
@@ -312,7 +313,7 @@ public class GiniVisionActivity extends Activity
         // analysis didn't complete or
         // the user rotated the image
         getSingleDocumentAnalyzer().removeListener();
-        showFragment(getAnalysisFragment(document), R.string.title_analysis);
+        showFragment(getAnalysisFragment(document));
     }
 
     @Override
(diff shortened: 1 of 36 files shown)


componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
```diff
@@ -163,14 +166,64 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            } else {
+                showFragment(getAnalysisFragment(document), R.string.title_review);
+            }
+        } catch (ImportedFileValidationException e) {
+            e.printStackTrace();
+            String message = "File cannot be analyzed";
+            if (e.getValidationError() != null) {
+                switch (e.getValidationError()) {
+                    case TYPE_NOT_SUPPORTED:
+                        message = "File type not supported.";
```
kikettas: This messages should be extracted to a `strings` file and written in german, right?

componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionAppCompatActivity.java
```diff
@@ -1,11 +1,15 @@
 package net.gini.android.vision.component;
```
kikettas: There is a lot of duplicated code in this file if you compare it with `GiniVisionActivity`. Is there a way to join them  and get rid of the duplicated code?

ginivision/src/main/java/net/gini/android/vision/GiniVision.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        return giniVisionIntent;
+    }
+
+    @NonNull
+    public static Document createDocumentForImportedFile(@NonNull final Intent intent,
+            @NonNull final Context context) throws ImportedFileValidationException {
+        final Uri uri = IntentHelper.getUri(intent);
+        if (uri == null) {
+            throw new ImportedFileValidationException("todo description no uri");
```
kikettas: I saw that `ImportedFileValidationException` stack trace is printed in other classes, so maybe we should add a proper description here.

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisActivity.java
```diff
@@ -164,6 +164,10 @@ public void noExtractionsFound() {
             startActivity(noResultsActivity);
             setResult(RESULT_NO_EXTRACTIONS);
             finish();
```
kikettas: This `finish()` call is duplicated and could be moved to the end of the method.

componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionAppCompatActivity.java
a-szotyori: Clients will have to study either one or the other depending on whether they use the support library or not. I would not abstract the code away for the example. I would like to keep everything in one file for each case. They share logic but shouldn't share code. Presenting an independent example for each case is more important than maintainability.

ginivision/src/main/java/net/gini/android/vision/GiniVision.java
a-szotyori: Whew, thanks for noticing this!

Issue #100, opened 2017-09-28
Feature toggle for document import


a-szotyori:
We need to support three cases:

1.  Default: Disabled
2. PDF only
3. PDF and Images

PR #102, opened 2017-09-29
Feature doc import toggle

a-szotyori:
Document import should be disabled by default. It also should be configurable to show only PDFs or PDFs and images.

### How to test
Run the example apps with different configurations: no DocumentImportFileTypes enum value, PDF and PDF_AND_IMAGES.

### Merging
Automatic.

### Ticket 
Issue #100


ginivision/src/main/java/net/gini/android/vision/DocumentImportFileTypes.java
```diff
@@ -0,0 +1,6 @@
+package net.gini.android.vision;
+
+public enum  DocumentImportFileTypes {
```
kikettas: Maybe the `enum` name was a bit confusing for me, because when I read it I thought it will have `PDF`, `IMAGES` as cases. 
Since this `enum` is used for the file import toggle and as a suggestion, could it be better to name it as something like _DocumentImportSupportedFileTypes_ or _DocumentImportEnabledFileTypes_? 
And also add a case when any type is supported, instead of using `null`.

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -312,13 +319,17 @@ private void initViews() {
...
-        if (FileChooserActivity.canChooseFiles(activity)) {
+        if (isDocumentImportEnabled() && FileChooserActivity.canChooseFiles(activity)) {
             mImportDocumentButtonEnabled = true;
             mButtonImportDocument.setVisibility(View.VISIBLE);
             showImportDocumentButtonAnimated();
         }
     }
 
+    private boolean isDocumentImportEnabled() {
+        return mDocumentImportFileTypes != null;
```
kikettas: Commented above.

ginivision/src/main/java/net/gini/android/vision/internal/fileimport/FileChooserActivity.java
```diff
@@ -172,6 +191,17 @@ public void onItemSelected(@NonNull final ProvidersAppItem item) {
                 }));
     }
 
+    private boolean shouldShowImageProviders() {
+        return mDocumentImportFileTypes != null
```
kikettas: Having the _NONE_ case this could be simplified.

Issue #104, opened 2017-10-05
Allow clients to check the imported document


a-szotyori:
In the Screen API clients should be able to check the imported document and return whether the library can proceed or not along with a message to be shown to the user.

In the Component API allow clients to notify the library that the document is not accepted and provide a message to be shown to the user.

PR #105, opened 2017-10-05
Feature client check document hook

a-szotyori:
Added a method for the clients to use to check imported documents.

Had to add loading the document content in the Camera Screen for it to be available to the clients.

Showing a loading indicator to the Camera Screen while the document content is loaded.

### How to test
Import documents and play around with the example apps.

### Merging
Automatic.

### Ticket 
Issue #104

### DIFF
--- a/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
+++ b/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
@@ -250,6 +250,23 @@ public class GiniVisionActivity extends Activity
     }
 
     @Override
+    public void onCheckImportedDocument(@NonNull final Document document,
+            @NonNull final DocumentCheckResultCallback callback) {
+        // We can apply custom checks here to an imported document and notify the Gini Vision
+        // Library about the result
+        // As an example we check the document size and allow documents smaller than 5MB
+        if (document.getData() != null) {
+            if (document.getData().length <= 5 * 1024 * 1024) {
+                callback.documentAccepted();
+            } else {
+                callback.documentRejected("Diese Datei ist leider größer als 5MB.");
+            }
+        } else {
+            callback.documentRejected(getString(R.string.gv_document_import_error));
+        }
+    }
+
+    @Override
     public void onDocumentReviewedAndAnalyzed(@NonNull Document document) {
         LOG.debug("Reviewed and analyzed document {}", document);
         // If we have received the extractions while in the Review Screen we don't need to go to
--- a/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
+++ b/componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
@@ -7,6 +7,7 @@ import android.app.AlertDialog;
 import android.app.Fragment;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
 import android.os.Handler;
 import android.support.annotation.NonNull;
@@ -28,6 +29,8 @@ import net.gini.android.vision.analysis.AnalysisFragmentListener;
 import net.gini.android.vision.analysis.AnalysisFragmentStandard;
 import net.gini.android.vision.camera.CameraFragmentListener;
 import net.gini.android.vision.camera.CameraFragmentStandard;
+import net.gini.android.vision.internal.util.IntentHelper;
+import net.gini.android.vision.internal.util.UriHelper;
 import net.gini.android.vision.noresults.NoResultsFragmentListener;
 import net.gini.android.vision.noresults.NoResultsFragmentStandard;
 import net.gini.android.vision.onboarding.OnboardingFragmentListener;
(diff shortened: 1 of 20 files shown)


componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
```diff
@@ -250,6 +250,23 @@ public void onDocumentAvailable(@NonNull Document document) {
...
+        // Library about the result
+        // As an example we check the document size and allow documents smaller than 5MB
+        if (document.getData() != null) {
+            if (document.getData().length <= 5 * 1024 * 1024) {
+                callback.documentAccepted();
+            } else {
+                callback.documentRejected("Diese Datei ist leider größer als 5MB.");
+            }
+        } else {
+            callback.documentRejected(getString(R.string.gv_document_import_error));
```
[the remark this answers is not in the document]
a-szotyori: Yes, it would be nicer, I'll extract the strings. :)

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -547,26 +561,111 @@ private void importDocumentFromIntent(final Intent data) {
...
+    private void disableInteraction() {
+        if (mCameraPreview == null
+                || mButtonImportDocument == null
+                || mButtonCameraTrigger == null) {
+            return;
+        }
+        mCameraPreview.setEnabled(false);
+        mButtonImportDocument.setEnabled(false);
+        mButtonCameraTrigger.setEnabled(false);
+    }
```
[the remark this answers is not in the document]
a-szotyori: I used to do methods with flag arguments, but then Rico pointed me to a section in Clean Code (he literally placed the book on my desk with a pan at the Flag Arguments section 😁): 
> Flag arguments are ugly. Passing a boolean into a function is a truly terrible practice. It
immediately complicates the signature of the method, loudly proclaiming that this function does more than one thing. It does one thing if the flag is true and another if the flag is false!
> 
> In Listing 3-7 we had no choice because the callers were already passing that flag
in, and I wanted to limit the scope of refactoring to the function and below. Still, the
method call render(true) is just plain confusing to a poor reader. Mousing over the call
and seeing render(boolean isSuite) helps a little, but not that much. We should have
split the function into two: renderForSuite() and renderForSingleTest().

It's a matter of preference for state toggling methods. I prefer two clear verbose methods. The thing about forgetting to add it to both methods can happen anyhow in case you have to do more than just pass a boolean to the view... and when having to toggle the state of many views we can add them to a list for example.

componentapiexample/src/main/java/net/gini/android/vision/component/GiniVisionActivity.java
```diff
@@ -249,23 +252,47 @@ public void onDocumentAvailable(@NonNull Document document) {
...
+            final Intent intent = document.getIntent();
+            if (intent == null) {
+                callback.documentRejected(getString(R.string.gv_document_import_error));
+                return;
+            }
+            final Uri uri = IntentHelper.getUri(intent);
+            if (uri == null) {
+                callback.documentRejected(getString(R.string.gv_document_import_error));
+                return;
+            }
```
kikettas: Die JAVA die!! 🔪

screenapiexample/src/main/java/net/gini/android/vision/screen/CameraScreenApiActivity.java
```diff
@@ -1,27 +1,159 @@
...
+    private boolean isDecodableToBitmap(final InputStream inputStream)
+            throws IOException {
+        BitmapFactory.Options options = new BitmapFactory.Options();
+        final Bitmap bitmap = BitmapFactory.decodeStream(inputStream, new Rect(), options);
+        return bitmap != null;
+    }
+
+    private boolean isPDF(final Uri uri, final byte[] magicBytes, final int read) {
+        boolean isRenderablePdf = true;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
kikettas: Since Android KitKat or below does not have PDFRenderer class, how do you know that the PDF is always renderable? Or is it only for the custom check?

screenapiexample/src/main/java/net/gini/android/vision/screen/CameraScreenApiActivity.java
```diff
@@ -1,27 +1,159 @@
...
+        }
+        if (fileDescriptor == null) {
+            LOG.error("Pdf not found");
+            return false;
+        }
+        try {
+            final PdfRenderer pdfRenderer = new PdfRenderer(fileDescriptor);
+            pdfRenderer.close();
+            return true;
+        } catch (IOException e) {
```
kikettas: Why don't you combine both try-catch?

screenapiexample/src/main/java/net/gini/android/vision/screen/CameraScreenApiActivity.java
a-szotyori: If I would return false by default then no PDF would be accepted on KitKat or below so I do the optimist. And it's just an example...

screenapiexample/src/main/java/net/gini/android/vision/screen/CameraScreenApiActivity.java
a-szotyori: To keep the exception catching close to the related throwing method. Otherwise it wouldn't be obvious which method throws which exception...

PR #119, opened 2017-11-10
Bugfix for PDF rendering on the Huawei MediaPad T2 10" Pro

a-szotyori:
The bug was cause by using different PdfRenderer instances on different threads. PdfRenderer is not thread safe and even though instances were not shared between threads the underlying libpdfium.so caused SIGSEGV faults. This issue was encountered only on this Huawei tablet.

To fix it we use one RendererLollipop instance in the AnalysisFragment which creates PdfRenderer instances on secondary threads (AsyncTask) and uses synchronization to allow only one at a time. The single RendererLollipop instance is the lock for synchronization.

### How to test
Run on the Huawei tablet and analyze different pdfs.

### Merging
Automatic.

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererPreLollipop.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererPreLollipop.java
@@ -1,7 +1,9 @@
 package net.gini.android.vision.internal.pdf;
 
+import android.graphics.Bitmap;
 import android.support.annotation.NonNull;
 
+import net.gini.android.vision.internal.AsyncCallback;
 import net.gini.android.vision.internal.util.Size;
 
 /**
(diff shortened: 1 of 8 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererPreLollipop.java
```diff
@@ -11,12 +13,18 @@
...
+        asyncCallback.onSuccess(null);
+    }
+
+    @Override
+    public void getPageCount(@NonNull final AsyncCallback<Integer> asyncCallback) {
+        asyncCallback.onSuccess(0);
     }
 
     @Override
     public int getPageCount() {
```
kikettas: Does this method make sense anymore? (Duplicated)

a-szotyori: It's the syncrhonous version of the other one. We need it when we do the pdf page count check in the [FileImportValidator](https://github.com/gini/gini-vision-lib-android/blob/87c3213d40b7dd6eecb0f341cfed5de58b7350f8/ginivision/src/main/java/net/gini/android/vision/internal/util/FileImportValidator.java#L101).

Issue #122, opened 2017-11-28
Unable to read file size for "file://" scheme


a-szotyori:
Some apps don't share the file using the [FileProvider](https://developer.android.com/reference/android/support/v4/content/FileProvider.html). Instead the file's absolute path is used with the "file://" scheme.

PR #123, opened 2017-11-28
Fixed retrieving file size for "file://" scheme Uris

a-szotyori:
Some apps share files by using a file Uri. For these querying the ContentResolver doesn't work and we have to use File to get the size and the filename.

### How to test
Run `UriHelperTest` and install one of the example apps on a device or emulator and send a file to the app with "open with".

### Merging
Nothing special.

### Ticket 
#122

### DIFF
--- a/ginivision/src/androidTest/java/net/gini/android/vision/util/UriHelperTest.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/util/UriHelperTest.java
@@ -2,12 +2,11 @@ package net.gini.android.vision.util;
 
 import static com.google.common.truth.Truth.assertThat;
 
-import static net.gini.android.vision.test.Helpers.deleteFileOrFolder;
-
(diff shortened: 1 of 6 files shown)


ginivision/src/androidTest/java/net/gini/android/vision/util/UriHelperTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private static final String TEST_FILE = "invoice.jpg";
+
+    @BeforeClass
+    public static void setUpClass() throws Exception {
+        setUpFileProvider();
+    }
+
+    private static void setUpFileProvider() throws IOException {
+        final File fileProviderDir = getFileProviderDir();
```
kikettas: Since you are using the file provider dir in your test cases and it is always the same, would it make sense to create an instance property for that and initialize it here?

PR #125, opened 2017-12-05
Refactor Component API Example

a-szotyori:
Separated the single activity with multiple fragments into separate activities for each fragment. This separation is a more likely scenario in a production environment and it also shows the different screens we have in the library.

Also ported the runtime permission requests from the showcase app and the safer gini api client credential injection (no need to add them to a git tracked file anymore).

### How to test
Run the component api example app.

### Merging
Automatic.


componentapiexample/src/main/java/net/gini/android/vision/component/camera/BaseCameraScreenHandler.java
```diff
@@ -0,0 +1,357 @@
...
+        // Delay notifying the coordinator to allow the camera fragment view to be created
+        new Handler().post(new Runnable() {
+            @Override
+            public void run() {
+                mGiniVisionCoordinator.onCameraStarted();
+            }
+        });
+    }
+
+    protected abstract CameraFragmentInterface showCameraFragment();
```
kikettas: Does not the method name represent the action? Maybe instead of _show_ it could be _get_ or _retrieve_.
I say this because 10 lines above you are assigning it to a variable, I know that you are returning the fragment after showing it but maybe it can be confusing.

componentapiexample/src/main/java/net/gini/android/vision/component/camera/BaseCameraScreenHandler.java
```diff
@@ -0,0 +1,357 @@
...
+        mCameraFragmentInterface.hideInterface();
+        setTitlesForOnboarding();
+        showOnboardingFragment();
+    }
+
+    protected abstract void showOnboardingFragment();
+
+    protected abstract void setTitlesForOnboarding();
+
+    private void startGiniVisionLibraryForImportedFile(final Intent importedFileIntent) {
```
kikettas: Since `GiniVisionFileImport.createDocumentForImportedFile` method only acepts `@NonNull`intents, I think that it would be nice to add the annotation also in this method.

componentapiexample/src/main/java/net/gini/android/vision/component/camera/standard/CameraScreenHandler.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    protected void setUpActionBar() {
+
+    }
+
+    @Override
+    protected CameraFragmentInterface showCameraFragment() {
+        mCameraFragment = CameraFragmentStandard.createInstance(
+                getGiniVisionFeatureConfiguration());
+        getActivity().getFragmentManager().beginTransaction()
```
kikettas: That is what I meant, you are showing it and creating it in a showing method. Maybe adding adding a _final get_ method for the private property `mCameraFragment` could be enough.

componentapiexample/src/main/java/net/gini/android/vision/component/noresults/compat/NoResultsScreenHandlerAppCompat.java
```diff
@@ -0,0 +1,68 @@
...
+        .commit();
+    }
+
+    @Override
+    protected void setTitles() {
+        final ActionBar actionBar = mAppCompatActivity.getSupportActionBar();
+        if (actionBar == null) {
+            return;
+        }
+        actionBar.setTitle("Keine Ergebnisse");
```
kikettas: Since this strings are used in more classes, could be useful to extract them!

componentapiexample/src/main/java/net/gini/android/vision/component/review/standard/ReviewScreenHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        return mReviewFragment;
+    }
+
+    @Override
+    protected void setTitles() {
+        final ActionBar actionBar = getActivity().getActionBar();
+        if (actionBar == null) {
+            return;
+        }
+        actionBar.setTitle("Seite überprüfen");
```
kikettas: Same with this string.

exampleShared/java/net/gini/android/vision/example/RuntimePermissionHandler.java
```diff
@@ -0,0 +1,276 @@
...
+ * Created by Alpar Szotyori on 05.12.2017.
+ *
+ * Copyright (c) 2017 Gini GmbH.
+ */
+
+public class RuntimePermissionHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(RuntimePermissionHandler.class);
+
+    private final Activity mActivity;
```
kikettas: I'm doubting here. Could it be a problem if this property is `strong`?

exampleShared/java/net/gini/android/vision/example/RuntimePermissionHandler.java
```diff
@@ -0,0 +1,276 @@
...
+
+    private void showAppDetailsSettingsScreen() {
+        final Intent intent = new Intent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
+        final Uri uri = Uri.fromParts("package", mActivity.getPackageName(), null);
+        intent.setData(uri);
+        mActivity.startActivity(intent);
+    }
+
+    private void showCameraPermissionRationale(final PermissionToken token) {
+        final android.support.v7.app.AlertDialog
```
kikettas: Why here the `AlertDialog` is from the support library but some in the other method it isn't?

componentapiexample/README.md
```diff
@@ -3,51 +3,64 @@ Component API Example App
...
 The `GiniVisionCoordinator` is used to show the onboarding on the first run. After the camera was started `GiniVisionCoordinator#onCameraStarted()` is called and, if this is the first run, the `OnboardingFragment` is shown in `GiniVisionActivity#onShowOnboarding()`.
 
-OnboardingFragment
-------------------
+Camera Screen
+-------------
 
-When the user clicks the onboarding menu item or the app starts the first time the `OnboardingFragment` is shown above the `CameraFragment`. When the user reached the end of the onboarding
-the `OnboardingFragment` is removed in `GiniVisionActivity#onCloseOnboarding()`. 
+The `BaseCameraScreenHandler` contains the main logic. The Activity and concrete handler in the `compat` and `standard` sub-packages provide the implementations dependent on whether the standard or Support Library components are used.
```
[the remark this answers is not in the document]
a-szotyori: I'll rephrase it to: 
The `compat` and `standard` sub-packages provide the implementations for the Support Library components and standard components, respectively. They contain an Activity and a concrete handler.

componentapiexample/README.md
```diff
@@ -3,51 +3,64 @@ Component API Example App
...
+The `BaseReviewScreenHandler` contains the main logic. The Activity and concrete handler in the `compat` and `standard` sub-packages provide the implementations dependent on whether the standard or Support Library components are used.
 
-When the document analysis successfully completed it calls `ReviewFragment#onDocumentAnalyzed()`. When the user clicks the next button either `GiniVisionActivity#onDocumentReviewedAndAnalyzed()` or `GiniVisionActivity#onProceedToAnalysisScreen()` is called. In the first method the extractions are shown in the `ExtractionsActivity` and in the second one the `AnalysisFragment` is shown.
+When the `ReviewFragment` starts the `ReviewFragmentListener#onShouldAnalyzeDocument()` is called where the document analysis is started while the user reviews the picture. If the picture was rotated the `BaseReviewScreenHandler` cancels the analysis in `ReviewFragmentListener#onDocumentWasRotated()`. If the analysis fails it caches an error message which will be passed to the Analysis Screen to be shown to the user.
```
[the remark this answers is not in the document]
a-szotyori: I rephrased it. What do you think?
The `ReviewFragmentListener#onShouldAnalyzeDocument()` method is called when the `ReviewFragment` starts. In this method the document analysis is started to use the time for analysis while the user reviews the picture. If the picture was rotated the `BaseReviewScreenHandler` cancels the analysis in `ReviewFragmentListener#onDocumentWasRotated()`. If the analysis fails it caches an error message which will be passed to the Analysis Screen to be shown to the user.

componentapiexample/src/main/java/net/gini/android/vision/component/analysis/standard/AnalysisScreenHandler.java
```diff
@@ -0,0 +1,69 @@
...
+        final ActionBar actionBar = getActivity().getActionBar();
+        if (actionBar == null) {
+            return;
+        }
+        actionBar.setTitle("");
+        actionBar.setSubtitle("Einen Moment bitte ...");
+    }
+
+    @Override
+    protected void setUpActionBar() {
```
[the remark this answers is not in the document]
a-szotyori: The standard way is that the ActionBar is managed for the Activity by the system, so no setting up is needed.

exampleShared/java/net/gini/android/vision/example/RuntimePermissionHandler.java
a-szotyori: I really appreciate that you point out these possible leaks! 👍 

In this case it's not an issue because the permission handler is owned by the activity only. The activity reference is passed only to Dexter and AlertDialogs. I've seen in Dexter's source, that they use weak references.

exampleShared/java/net/gini/android/vision/example/RuntimePermissionHandler.java
a-szotyori: Auto-import... saves time, but soooo easy to import something unintended....

PR #130, opened 2017-12-12
Payment QRCode Detection and Parsing

a-szotyori:
**Please Note**: Tests will be added later this week. It would be great if we can release this as a beta version without tests to be able to use it in the showcase app.

BezahlCode and EPC069-12 (Stuzza (AT), GiroCode (DE)) is detected using Google's Mobile Vision API. Parsing is done according to the format specifications with IBAN validation and amount+currency normalization to fit the [Gini API format for amount](http://developer.gini.net/gini-api/html/entity_reference.html#extraction-entity-amount). 

### How to test
Run the example apps and scan invoices/remittance slips with the supported QRCode formats.

### Merging
Automatic

### Ticket 
Issues #126, #127


ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
```diff
@@ -605,6 +606,11 @@ public void onDocumentAvailable(@NonNull Document document) {
         }
     }
 
+    @Override
+    public void onPaymentDataAvailable(@NonNull final PaymentData paymentData) {
+
```
kikettas: This method should be implemented, right?

ginivision/src/main/java/net/gini/android/vision/internal/qrcode/AmountAndCurrencyNormalizer.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+        try {
+            BigDecimal amountBigDecimal = new BigDecimal(amount.replace(",", "."));
+            if (amountBigDecimal.scale() < 2) {
+                amountBigDecimal = amountBigDecimal.setScale(2, BigDecimal.ROUND_HALF_EVEN);
+            }
+            return amountBigDecimal.toPlainString() + ":" + currency;
+        } catch (final NumberFormatException ignored) {
+        }
+        return "";
```
kikettas: You can simplify this a little bit. 
```
if !TextUtils.isEmpty(amount) {
try ...
} else {
      return ""
}```

File: ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorImpl.java [hunk 3]
```diff
@@ -0,0 +1,71 @@
...
+        mHandlerThread = new HandlerThread("QRCodeDetectorThread",
+                Process.THREAD_PRIORITY_BACKGROUND);
+        mHandlerThread.start();
+        mHandler = new QRCodeDetectorHandler(mHandlerThread.getLooper(), qrCodeDetectorTask);
+    }
+
+    @Override
+    public void detect(@NonNull final byte[] image, @NonNull final Size imageSize,
+            final int rotation) {
+        if (mListener == null) {
```
kikettas: Here is `mHandler` not `mListener`, right?

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentListener.java
```diff
@@ -20,6 +21,8 @@
      */
     void onDocumentAvailable(@NonNull Document document);
```
kikettas: Don't forget to add the documentation before the final release!

ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
a-szotyori: Yes, I will add to the documentation that this has to be overriden in a subclass to get the payment data with the Screen API.

ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorImpl.java
a-szotyori: If there is no Listener I avoid processing any images. (If no one is interested why do the work? 😄). The handler cannot be null.

I will add a comment to clarify this.

PR #132, opened 2017-12-14
Feature qr code detection improvements

a-szotyori:
* BarcodeDetector isOperational check is tried 4 times with 500ms delays between retries and if it fails only a warning is logged and detection is disabled.
* Show QRCode detected popup only when onboarding is not shown and document upload hint is not shown.
* Hide and show QRCode detected popup, if different QRCode is detected.

Issue #127

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorTask.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorTask.java
@@ -21,8 +21,15 @@ public interface QRCodeDetectorTask {
     List<String> detect(@NonNull final byte[] image, @NonNull final Size imageSize,
             final int rotation);
 
-    boolean isOperational();
+    void isOperational(@NonNull final Callback callback);
 
     void release();
+
+    public interface Callback {
+
+        void onResult(final boolean isOperational);
+
+        void onInterrupted();
+    }
 
 }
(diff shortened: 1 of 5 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorTask.java
```diff
@@ -21,8 +21,15 @@
     List<String> detect(@NonNull final byte[] image, @NonNull final Size imageSize,
             final int rotation);
 
-    boolean isOperational();
+    void isOperational(@NonNull final Callback callback);
```
kikettas: IMO the name of this method is a little bit confusing now that is returning a void. Maybe `checkAvailability` or similar could be more precise.

ginivision/src/main/java/net/gini/android/vision/internal/qrcode/QRCodeDetectorTaskGoogleVision.java
```diff
@@ -78,13 +102,44 @@ private String barcodesToString(final SparseArray<Barcode> barcodes) {
         return builder.toString();
     }
 
-    @Override
-    public boolean isOperational() {
-        return mBarcodeDetector.isOperational();
-    }
+    class IsOperationalRetryRunnable implements Runnable {
```
kikettas: Same here, for me at least is not clear.

PR #134, opened 2018-01-10
QR Code Feature Documentation

a-szotyori:
Wrote a documentation for the QR Code feature.

Stephan (@EarlOfEgo) if you have time we would be very grateful if you could take a close look at the Google Play Services section. I don't have much experience with it and maybe you know of some additional caveats we can point out for our clients. Thank you!

#129


ginivision/src/doc/source/updating-to-2-5-0.rst
```diff
@@ -0,0 +1,83 @@
...
+With version 2.5.0 we added the Google Mobile Vision API dependency which is part of the Google Play Services. We only depend on the ``play-services-vision`` package.
+
+In case you already use the Google Mobile Vision API there might be a conflict related to the ``<meta-data android:name="com.google.android.gms.vision.DEPENDENCIES">`` tag after updating to 2.5.0. If this is the case add the attribute ``tools:replace="android:value"``. Make sure that the value contains ``barcode``:
+
+.. code-block:: xml
+
+    <meta-data
+        android:name="com.google.android.gms.vision.DEPENDENCIES"
+        tools:replace="android:value"
+        android:value="face,barcode"/>
```
kikettas: The `face` value is not needed.

ginivision/src/doc/source/updating-to-2-5-0.rst
```diff
@@ -0,0 +1,83 @@
...
+    // Enable QR Code scanning
+    final GiniVisionFeatureConfiguration giniVisionFeatureConfiguration =
+            GiniVisionFeatureConfiguration.buildNewConfiguration()
+                    .setQRCodeScanningEnabled(true)
+                    .build();
+    CameraFragmentCompat.createInstance(giniVisionFeatureConfiguration);
+
+.. important::
+
+    When your application is installed Google Mobile Services will download libraries to the device in order to do QR Code detection. Under certain circumstances (user not online, slow connection or lack of sufficient storage space) the libraries will not be ready at the time your app starts the Camera Screen and QR Code detection will be silently disabled until the next time the Camera Screen starts.
```
EarlOfEgo: Afaik it is only downloaded ones per device, so if another app already used it it should be there. Maybe point that out as well.

a-szotyori:
Thanks for the review! I integrated your suggestions.

Issue #135, opened 2018-01-11
Expose QR Code and parsed payment data as a json string


a-szotyori:
To track QR Codes we have to expose it together with the parsed payment data as a json string. This json can be then be uploaded with the Gini API to create a document.

### Json format (draft)
```
{
    "qrcode": "<raw qr code string>"
    "paymentdata": {
        "amountToPay": "<(optional) amount in Gini API format: 25.79:EUR>"
        "paymentRecipient": "<(optional) recipient name>"
        "iban": "<(mandatory) iban>"
        "bic": "<(optional) bic>"
        "paymentReference": "<(optional) reference>"
    }
}

If any of the labels is missing, don't add it with an empty value to the json.
```

ricogini:
I would prefer to have the complete labels optional. If a label was not extracted it should not be included in the json.

PR #139, opened 2018-01-12
Added method to write PaymentData with unparsed QR Code to a json

a-szotyori:
### How to test
Run the tests in PaymentDataTest.

### Ticket 
#135

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/PaymentData.java
+++ b/ginivision/src/main/java/net/gini/android/vision/PaymentData.java
@@ -3,6 +3,13 @@ package net.gini.android.vision;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.text.TextUtils;
+import android.util.JsonWriter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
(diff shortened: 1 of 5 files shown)


ginivision/src/main/java/net/gini/android/vision/PaymentData.java
```diff
@@ -90,28 +137,34 @@ public boolean equals(final Object o) {
...
-        if (!mBIC.equals(that.mBIC)) {
+        if (mBIC != null ? !mBIC.equals(that.mBIC) : that.mBIC != null) {
             return false;
         }
-        if (!mIBAN.equals(that.mIBAN)) {
+        if (mIBAN != null ? !mIBAN.equals(that.mIBAN) : that.mIBAN != null) {
             return false;
         }
-        if (!mPaymentRecipient.equals(that.mPaymentRecipient)) {
+        if (mPaymentRecipient != null ? !mPaymentRecipient.equals(that.mPaymentRecipient)
```
kikettas: I would extract this to a separated method, it is more clear and less prone to errors.

a-szotyori: equals() and hashCode() are boilerplate generated methods. I would leave them as they are.

Issue #137, opened 2018-01-11
Update example apps and documentation to upload QR Codes to Gini


a-szotyori:
We will require from our clients to upload the QR Code with the payment data. We will also need them to send us feedback for the QR Code.

This can be achieved by creating a Gini document with the QR Code and payment data json (#135) and sending feedback for that document.

PR #140, opened 2018-01-15
Enable uploading of QR Codes to the Gini API

a-szotyori:
To track QR Codes we had to change our beta QR Code public api quite a bit. We don't expose the payment data anymore directly. We create a json containing the raw QR Code string and the payment data we parsed from it (#135). This json is publicly available as an utf-8 encoded byte array which should be uploaded to the Gini API to get the payment data from the QR Code. Feedback should also be sent for it.

The example apps were updated to upload the QR Code. Documentation was also updated.

**Note:**
The previous beta QR Code public api used the term `Payment Data` to cover future cases where other sources than QR Codes are used to read payment data (e.g. NFC). The requirement to upload the QR Code required us to drop that naming and mention QR Codes explicitly for clarity and convenience. That's why there are so many changed files.

### How to test
Run the example apps and scan BezahlCode, Stuzza or GiroCode.

### Ticket 
Issue #137

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
+++ b/ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
@@ -152,26 +152,26 @@ import java.util.ArrayList;
  *         gvCustomFont} with the path to the font file in your {@code assets} folder
  *     </li>
  *     <li>
- *         <b>Payment data (QRCode) detected popup background:</b> via the color resource named {@code gv_payment_data_detected_popup_background}
+ *         <b>QRCode detected popup background:</b> via the color resource named {@code gv_qrcode_detected_popup_background}
  *     </li>
  *     <li>
- *         <b>Payment data (QRCode) detected popup texts:</b> via the string resources named {@code gv_payment_data_detected_popup_message_1} and
- *         {@code gv_payment_data_detected_popup_message_2}
+ *         <b>QRCode detected popup texts:</b> via the string resources named {@code gv_qrcode_detected_popup_message_1} and
+ *         {@code gv_qrcode_detected_popup_message_2}
  *     </li>
  *     <li>
- *         <b>Payment data (QRCode) detected popup text sizes:</b>  via overriding the styles named {@code
- *         GiniVisionTheme.Camera.PaymentDataDetectedPopup.Message1.TextStyle} and {@code
- *         GiniVisionTheme.Camera.PaymentDataDetectedPopup.Message2.TextStyle} and setting an item named {@code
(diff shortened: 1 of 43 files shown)


ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentInterface.java
```diff
@@ -79,4 +81,29 @@
...
+    /**
+     * <p>
+     *     Call this method to hide the activity indicator and enable user interaction.
+     * </p>
+     */
+    void hideActivityIndicatorAndEnableInteraction();
+
+    /**
+     * <p>
+     *     Call this method when to show an error message to the user in the Camera Screen.
```
kikettas: ~when~

ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
```diff
@@ -297,7 +297,8 @@
  * </ul>
  * </p>
  **/
-public class CameraActivity extends AppCompatActivity implements CameraFragmentListener {
+public class CameraActivity extends AppCompatActivity implements CameraFragmentListener,
+        CameraFragmentInterface {
```
kikettas: I'm not really sure about this interface name. When I have seen it, I have thought about a generic java `interface`, not as an interface related to user interface.

screenapiexample/src/main/java/net/gini/android/vision/screen/CameraScreenApiActivity.java
```diff
@@ -161,12 +174,27 @@ private boolean isRenderablePDF(final Uri uri) {
...
-        finish();
+    public void onQRCodeAvailable(@NonNull final QRCodeDocument qrCodeDocument) {
+        showActivityIndicatorAndDisableInteraction();
+        mSingleDocumentAnalyzer.cancelAnalysis();
+        mSingleDocumentAnalyzer.analyzeDocument(qrCodeDocument,
+                new DocumentAnalyzer.Listener() {
+                    @Override
+                    public void onException(final Exception exception) {
+                        hideActivityIndicatorAndEnableInteraction();
+                        showError("Could not use the QR Code. Try again or take a picture of your document.", 4000);
```
kikettas: Same here.

ginivision/src/main/java/net/gini/android/vision/camera/CameraActivity.java
a-szotyori: It is a generic interface and it contains only UI related methods now which might change in the future.

PR #143, opened 2018-01-29
Use Android Gradle Plugin 3.0.1

a-szotyori:
Updated to using the newest Android Gradle Plugin.

### How to test
Compile and run the example apps.


.idea/misc.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EarlOfEgo: Don't think this should be in git.

componentapiexample/build.gradle
```diff
@@ -61,25 +61,28 @@ repositories {
 }
 
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile deps.appCompatv7
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
```
EarlOfEgo: 👍

.idea/misc.xml
EarlOfEgo: It's even in the `.gitignore` file https://github.com/gini/gini-vision-lib-android/blob/master/.gitignore#L31
Imo the whole `.idea/` folder should be on it.

a-szotyori: I left it because according to IntelliJ it should be versioned https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

a-szotyori: Still strange why it wasn't ignored though... well I'll remove it. I remember now, that I ignored as it kept changing the jdk version.

EarlOfEgo: yes it should be versioned when:
>If you decide to share IDE project files with other developers, follow these guidelines

But my opinion is more against all related IDE files in version control, didn't make this clear in my comment.(However opinion based of course)

a-szotyori: I see your point. I think for Android projects as Android Studio is the standard IDE it's helpful to have project files in the repo.

Issue #149, opened 2018-03-09
Multi-Page Document from Multiple Imported Images


a-szotyori:
Allow importing of multiple images from the Camera Screen and using "Open with".

Show the images in the Multi-Page Review Screen.

PR #152, opened 2018-03-16
Feature multi page import ui

a-szotyori:
* Imported images are shown in the image stack
* Imported and taken images are added to the existing image stack (if there is one)
* PDF import is disabled when in multi-page mode
* Multi-page file import validation errors are shown in the Multi-Page Review Screen (UI and messages are not final)

### How to test
Run the Screen API example app and verify the changes above.

### Ticket 
Issue #149

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
+++ b/ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
@@ -56,6 +56,7 @@ import net.gini.android.vision.document.GiniVisionMultiPageDocument;
 import net.gini.android.vision.document.ImageDocument;
 import net.gini.android.vision.document.ImageMultiPageDocument;
 import net.gini.android.vision.document.QRCodeDocument;
+import net.gini.android.vision.internal.AsyncCallback;
 import net.gini.android.vision.internal.camera.api.CameraController;
 import net.gini.android.vision.internal.camera.api.CameraException;
 import net.gini.android.vision.internal.camera.api.CameraInterface;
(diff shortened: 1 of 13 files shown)


ginivision/src/main/java/net/gini/android/vision/document/GiniVisionDocumentError.java
```diff
@@ -0,0 +1,51 @@
...
+import android.os.Parcelable;
+import android.support.annotation.NonNull;
+
+/**
+ * Created by Alpar Szotyori on 15.03.2018.
+ *
+ * Copyright (c) 2018 Gini GmbH.
+ */
+
+public class GiniVisionDocumentError implements Parcelable {
```
kikettas: Is this a real Error class or just a string wrapper? Since it there could be different error cases, would it make sense to have this as an `enum`?

ginivision/src/main/java/net/gini/android/vision/camera/CameraFragmentImpl.java
```diff
@@ -974,25 +992,77 @@ private void createMultiPageDocumentAndCallListener(@NonNull final Context conte
...
+    private void addMultiPageDocumentError(final String string) {
+        final ImageDocument document = DocumentFactory.newEmptyImageDocument();
+        mMultiPageDocument.addDocument(document);
+        final GiniVisionDocumentError documentError = new GiniVisionDocumentError(string);
+        mMultiPageDocument.addErrorForDocument(document, documentError);
+    }
+
+    private Bitmap getBitmap(final ImageDocument imageDocument) {
+        final Photo photo = PhotoFactory.newPhotoFromDocument(imageDocument);
+        // TODO: get rid of bitmap rotation -> rotate only the ImageView in the stack
```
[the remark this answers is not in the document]
a-szotyori: Yes, I promise. Slap me if you still find it like this in the released version. 😄

ginivision/src/main/java/net/gini/android/vision/document/GiniVisionDocumentError.java
a-szotyori: I used a class to have flexibility considering the error payload. Now it's only a string and when adding networking there might be a need for some additional information. It can also be that we use a String or an Enum.

Issue #148, opened 2018-03-09
Multi-Page Document Memory Management


a-szotyori:
Store images on disk and keep a limited amount of bitmaps in memory.

PR #153, opened 2018-03-22
Feature multi page memory management

a-szotyori:
Loading documents' data and creating previews are handled by memory cache managers backed by an LRU cache. The size of the LRU cache is determined by the available memory (kudos to @EarlOfEgo and the Switch SDK team). The memory cache managers are called `DocumentDataMemoryCache` and `PhotoMemoryCache`.

Storing images on disk is handled by the `ImageDiskStore` class. This writes byte arrays to the app's internal folder and returns a`Uri` which is then used in documents to read their data from.

### How to test
Run the Screen API Example App and take/import a lot of images and see if the app crashes.

### Ticket 
Issue #148


a-szotyori:
Sorry, I will remove the `// WIP-MM` comments

Issue #154, opened 2018-03-22
Use Copies of Imported Images


a-szotyori:
In order to modify imported images we have to copy them into the app's own storage space and work with the copy.

Images for multi-page documents are stored temporarily on the disk to enable unloading them from memory and loading them when required. This means that imported images cannot be modified only in-memory anymore as before. They can be flushed from memory loosing any modifications.

PR #156, opened 2018-03-26
Feature multi page copy imported images

a-szotyori:
Imported images are copied to the apps internal storage to be editable.

### How to test
Run the Screen API Example App and import images from the Camera Screen, rotate them and verify that the images in the image stack are rotated accordingly.
Also import images with "open with". In this case you can check that copies exist and rotation was applied using Android Studio's Device File Explorer (or with adb using run-as to copy the images from the app internal storage to the SDCard).

### Ticket 
Issue #154

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
+++ b/ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
@@ -237,11 +237,15 @@ class Exif {
 
         @NonNull
         public Builder setOrientationFromDegrees(final int degrees) {
-            final byte[] bytes = new byte[1];
-            bytes[0] = (byte) rotationToExifOrientation(degrees);
-            final TiffOutputField orientationOutputField = new TiffOutputField(
-                    TiffTagConstants.TIFF_TAG_ORIENTATION, FieldType.SHORT, 1, bytes);
-            mIfd0Directory.add(orientationOutputField);
+            try {
+                final short orientation = rotationToExifOrientation(degrees);
+                final byte[] bytes = FieldType.SHORT.writeData(orientation,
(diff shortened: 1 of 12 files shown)


ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/Exif.java
```diff
@@ -237,11 +237,15 @@ public Builder setUserComment(final String userComment) {
...
-            final byte[] bytes = new byte[1];
-            bytes[0] = (byte) rotationToExifOrientation(degrees);
-            final TiffOutputField orientationOutputField = new TiffOutputField(
-                    TiffTagConstants.TIFF_TAG_ORIENTATION, FieldType.SHORT, 1, bytes);
-            mIfd0Directory.add(orientationOutputField);
+            try {
+                final short orientation = rotationToExifOrientation(degrees);
+                final byte[] bytes = FieldType.SHORT.writeData(orientation,
+                        mTiffOutputSet.byteOrder);
+                final TiffOutputField orientationOutputField = new TiffOutputField(
```
kikettas: Weren't TIFF's unavailable in Android? Also, are you about leaving the catch closure empty?

a-szotyori: Exif uses Tiff tag information format for compatibility `Related  attribute  information  for  both  compressed  and  uncompressed  files  is  stored  in  the  tag  information  format  defined in TIFF Rev. 6.0` (Page 2 in http://www.exif.org/Exif2-2.PDF)

That exception is thrown by the `FieldType.SHORT.writeData()` method in case the type is not a short, which in this case cannot happen. I decided to ignore it.

Issue #175, opened 2018-04-06
Multi-Page Review Screen: make the upload error view on pages nicer


a-szotyori:
Currently a simple red text is shown centered on each page. The error view should look nice on pages with image previews and also on pages without.

Issue #170, opened 2018-04-06
Review Screen: Show page upload progress


a-szotyori:
Implement the Sketch design (#169) and possibly adapt it to the implementation needs.

Issue #168, opened 2018-04-06
Review Screen: Hide next button until all pages have been uploaded


a-szotyori:
We have to stay in the Review Screen until all page uploads have finished to be able to show upload errors for the failed pages.

PR #197, opened 2018-05-11
Feature multi page review upload progress

a-szotyori:
* Upload in progress and result shown on thumbnails.
* Upload error message shown on the document's preview page. (Design will be updated in #175)
* Next button hidden until all uploads were successful.

Screenshot

### How to test
Runt one of the example apps.

### Ticket 
Issues #170, #168

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
+++ b/ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
@@ -27,6 +27,7 @@ import net.gini.android.vision.Document;
 import net.gini.android.vision.GiniVision;
 import net.gini.android.vision.R;
 import net.gini.android.vision.document.GiniVisionDocument;
+import net.gini.android.vision.document.GiniVisionDocumentError;
 import net.gini.android.vision.document.ImageDocument;
 import net.gini.android.vision.document.ImageMultiPageDocument;
 import net.gini.android.vision.internal.AsyncCallback;
@@ -415,7 +416,9 @@ public class MultiPageReviewFragment extends Fragment implements MultiPageReview
                                         final Throwable throwable) {
                                     if (throwable != null &&
                                             !NetworkRequestsManager.isCancellation(throwable)) {
-                                        // WIP-MPA: show error for imageDocument on ViewPager page
+                                        final String errorMessage = getString(
+                                                R.string.gv_multi_page_review_upload_error);
+                                        showErrorOnPreview(errorMessage, imageDocument);
                                         mThumbnailsAdapter.setUploadState(
                                                 ThumbnailsAdapter.UploadState.FAILED,
                                                 imageDocument);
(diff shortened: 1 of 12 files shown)


ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
```diff
@@ -397,6 +397,10 @@ public void onStart() {
         if (!mPreviewsShown) {
             observeViewTree();
         }
+        uploadDocuments();
```
kikettas: Why don't you upload the documents when they are created instead of when multipage fragment starts?

ginivision/src/main/java/net/gini/android/vision/review/multipage/thumbnails/ThumbnailsAdapter.java
```diff
@@ -317,14 +346,35 @@ public void setItemTouchHelper(final ItemTouchHelper itemTouchHelper) {
...
+        }
+    }
+
     enum ViewType {
         THUMBNAIL,
         PLUS_BUTTON
     }
 
+    public enum UploadState {
+        NOT_STARTED,
```
kikettas: Does it make sense to have this state? I mean, the initial state should always be `IN_PROGRESS`

ginivision/src/main/java/net/gini/android/vision/review/multipage/thumbnails/ThumbnailsAdapter.java
```diff
@@ -353,23 +409,64 @@ void showActivityIndicator() {
...
+            uploadResultIconBackground.setVisibility(View.INVISIBLE);
+            uploadResultIconForeground.setVisibility(View.INVISIBLE);
         }
 
         boolean isDragAllowed() {
-            return mViewType == ViewType.THUMBNAIL;
+            return viewType == ViewType.THUMBNAIL;
+        }
+
+        void showUploadSuccess() {
```
kikettas: Could you combine this two methods into a switch? I'm fine with this but I think it is the perfect place to have a switch for the `UploadState`

ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
a-szotyori: To be in line with the single-page review screen I want to do uploads once the user is in the review screen.

I also want to keep upload out of the camera screen to not have to deal with passing upload related information (error, progress) to the review screen.

ginivision/src/main/java/net/gini/android/vision/review/multipage/thumbnails/ThumbnailsAdapter.java
a-szotyori: From the view's point of view it makes sense for me. The default state for it is NOT_STARTED to not make assumptions and it shows `IN_PROGRESS` only when the upload is really in progress.

ginivision/src/main/java/net/gini/android/vision/review/multipage/thumbnails/ThumbnailsAdapter.java
a-szotyori: Well they're invoked from within a switch. 😁 This ViewHolder doesn't know about the UploadState. These two methods allow the showing of upload success/failure. Under which circumstances they're invoked is imho irrelevant for the ViewHolder.

Issue #163, opened 2018-04-06
Review Screen: hide "+" button for imported files with "open with"


a-szotyori:
Even when multi-page is enabled the "+" button has to be hidden in the Review Screen. Sending files to GVL with "open with" doesn't start the Camera Screen so there's no way to add additional files currently.

PR #200, opened 2018-05-14
Feature always show multi page review and hide "+" button for "open with"

a-szotyori:
When multi-page is enabled (which it currently always is) it goes directly to the Multi-Page Review Screen.

Removed the "+" floating action button from the Single Page Review Screen.

Also the "+" is hidden for "open with" imported documents in the Multi-Page Review Screen.

### How to test
Run one of the example apps.

### Ticket 
Issue #199, #163

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
+++ b/ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
@@ -231,7 +231,7 @@ public class MultiPageReviewFragment extends Fragment implements MultiPageReview
         };
 
         mThumbnailsAdapter = new ThumbnailsAdapter(activity, mMultiPageDocument,
-                mThumbnailsAdapterListener);
+                mThumbnailsAdapterListener, shouldShowPlusButton());
         mThumbnailsRecycler.setAdapter(mThumbnailsAdapter);
 
         mThumbnailsScroller = new LinearSmoothScroller(activity);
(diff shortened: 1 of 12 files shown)


ginivision/src/main/java/net/gini/android/vision/review/multipage/MultiPageReviewFragment.java
```diff
@@ -248,6 +248,10 @@ public void onPlusButtonClicked() {
         mThumbnailsRecycler.setItemAnimator(itemAnimator);
     }
 
+    private boolean shouldShowPlusButton() {
+        return mMultiPageDocument.getImportMethod() != Document.ImportMethod.OPEN_WITH;
```
kikettas: Not sure about this. What if the user wants to add a page after importing one/several images?

a-szotyori: When launching with "open with" the user selected images in an app and sent it to GVL and there is no method to add more pages because we go directly to the Review Screen. More images can be added from within the GVL only in the Camera Screen.

The other import method is "picker" which is means the user was in the Camera Screen and used the file picker so we can show the "+" button. Also when the import method is not set the user must have come from the Camera Screen.

kikettas: And there is no way to add the camera to the stack when that occurs?

a-szotyori: Not without a major overhaul.

Issue #186, opened 2018-04-24
Camera Screen: Add a subtitle to the import button


a-szotyori:
We learned from the multi-page user testing that subtitles are needed to clarify what the icon is for.

Issue #183, opened 2018-04-24
Camera Screen: Add a subtitle to the image stack


a-szotyori:
A subtitle will help clarify what the image stack's purpose is.

PR #201, opened 2018-05-14
Feature camera screen icon subtitles (after #200)

a-szotyori:
device-2018-05-14-173848

### Ticket 
Issues #186, #183


ginivision/src/main/java/net/gini/android/vision/camera/ImageStack.java
```diff
@@ -165,13 +168,15 @@ public void setImages(@Nullable final List<Bitmap> bitmaps) {
         stackItem3.setBackgroundColor(Color.TRANSPARENT);
         badge.setText("");
         badge.setVisibility(INVISIBLE);
+        subtitle.setVisibility(INVISIBLE);
```
kikettas: Why not an `ifelse` for this? Something like: 
```
if (bitmaps == null || bitmaps.size() == 0) { 
// Here you make the badge and subtitle invisible
} else {
 // The remaining code
}
```

This way you would make it more clear and wouldn't have redundant code.

PR #209, opened 2018-05-24
Fixed QRCode crash caused by Indian IFSC QRCodes

a-szotyori:
Indian IFSC QRCodes start also with `bank:` and even though they don't contain a URI `Uri.parse()`parsed it and returned `bank` as the scheme so our parser didn't stop and tried to get query parameters which caused an `UnsupportedOperationException` in `Uri`. The exception is caught now and wrapped in the expected exception.

### How to test
Run one of the example apps and scan the QRCode below:
testqrcode


ginivision/src/main/java/net/gini/android/vision/internal/qrcode/BezahlCodeParser.java
```diff
@@ -35,19 +35,29 @@ public PaymentQRCodeData parse(@NonNull final String qrCodeContent)
...
         return new PaymentQRCodeData(qrCodeContent, paymentRecipient, paymentReference, iban, bic,
                 amount);
     }
+
+    private String getQueryParameter(@NonNull final Uri uri, @NonNull final String key) {
+        try {
+            return uri.getQueryParameter(key);
+        } catch (final UnsupportedOperationException e) {
+            throw new IllegalArgumentException(
+                    "QRCode content does not conform to the BerzahlCode format: "
```
kikettas: ~BerzahlCode~ BezahlCode

Issue #176, opened 2018-04-06
Move GiniVisionFileIimport methods to GiniVision


a-szotyori:
We need to unify the scattered public API in the new GiniVision class.

PR #210, opened 2018-05-25
Feature multi page file import (after #208)

a-szotyori:
* Made file import methods instance methods on `GiniVision`. The old methods will be deprecated in the next PR.
* Cleaned up file import code and eliminated duplicated logic.
* Fixed PDF "open with".

### How to test
Run the example apps to import images and PDFs.

### Ticket 
Issue #176


ginivision/src/main/java/net/gini/android/vision/AsyncCallback.java
```diff
@@ -0,0 +1,13 @@
+package net.gini.android.vision;
+
+/**
+ * Callback interface for asynchronous tasks.
+ */
+public interface AsyncCallback<T, E extends Exception> {
```
kikettas: 👍

ginivision/src/main/java/net/gini/android/vision/internal/fileimport/AbstractImportImageUrisAsyncTask.java
```diff
@@ -0,0 +1,214 @@
...
+                    }
+                    if (isCancelled()) {
+                        LOG.debug("Import cancelled for uri {}", uri);
+                        return null;
+                    }
+                    // Create compressed Document
+                    final ImageDocument compressedDocument =
+                            DocumentFactory.newImageDocumentFromPhoto(photo, localUri);
+                    LOG.debug("Compressed ImageDocument created from uri {}", uri);
+                    multiPageDocument.addDocument(compressedDocument);
```
kikettas: Since the point of this `if` is to add a document to the multipage document at the end, you could extract this huge clause to a method and then either return `null` if the document is `null` or add it to the multipage document.

PR #222, opened 2018-08-09
Bugfix protected pdfs

a-szotyori:
Pdfs with password caused a `SecurityException` when instantiating the `PdfRenderer` on Android API Levels 21+.

Added an additional file import validation when importing pdfs which checks whether the pdf is password protected. Also `SecurityExceptions` are gracefully handled now.

### How to test
A pdf with password and one without can be found in the androidTest assets folder: `ginivision/src/androidTest/assets`. Copy these to a device and open them in one of the example apps. The password protected pdf should cause a validation alert dialog to be shown.


ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererLollipop.java
```diff
@@ -78,10 +85,23 @@ private PdfRenderer getPdfRenderer() {
             return new PdfRenderer(fileDescriptor);
         } catch (final IOException e) {
             LOG.error("Could not read pdf", e);
+            closeFileDescriptor(fileDescriptor);
+        } catch (final SecurityException e) {
+            LOG.error("Could not read pdf", e);
+            closeFileDescriptor(fileDescriptor);
+            throw e;
```
EarlOfEgo: Why throwing the exception again if it is handled here? The only thing that is done [with it](https://github.com/gini/gini-vision-lib-android/pull/222/files#diff-50519542387040bda5205076d7e2d4d7R55) is to print the same error in the log.

ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererLollipop.java
```diff
@@ -130,7 +150,12 @@ public void onCancelled() {
 
     @Override
     public synchronized int getPageCount() {
-        final PdfRenderer pdfRenderer = getPdfRenderer();
+        PdfRenderer pdfRenderer = null;
+        try {
+            pdfRenderer = getPdfRenderer();
+        } catch (final SecurityException e) {
+            LOG.error("Could not read pdf", e);
```
EarlOfEgo: As stated above, there is no need to handle the exception twice.

ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererLollipop.java
```diff
@@ -139,6 +164,21 @@ public synchronized int getPageCount() {
...
     }
 
+    @Override
+    public boolean isPdfPasswordProtected() {
+        final PdfRenderer pdfRenderer;
+        try {
+            pdfRenderer = getPdfRenderer();
+        } catch (final SecurityException e) {
+            LOG.debug("PDF is password protected", e);
+            return true;
```
EarlOfEgo: Ok the exception is needed here, but I find it weird that a side effect of a method is used to determine the state of password protection.

ginivision/src/androidTest/java/net/gini/android/vision/internal/pdf/RendererLollipopTest.java
```diff
@@ -0,0 +1,110 @@
...
+public class RendererLollipopTest {
+
+    private static final String PDF = "invoice.pdf";
+    private static final String PDF_WITH_PASSWORD = "invoice-password.pdf";
+
+    private static Uri sPdfContentUri;
+    private static Uri sPdfWithPasswordContentUri;
+
+
+    @BeforeClass
```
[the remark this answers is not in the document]
a-szotyori: To copy the pdf files to the file provider directory only once. The tests don't modify the files - only read them via content provider URIs (`content://...`).

ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererLollipop.java
a-szotyori: To be able to catch the security exception in `isPdfPasswordProtected()`.

ginivision/src/main/java/net/gini/android/vision/internal/pdf/RendererLollipop.java
a-szotyori: I know. Would be great if `PdfRenderer` would have a static method to check whether a pdf is protected or not. I will look into dealing with the `SecurityException` in a cleaner way. I think I can refactor it a bit.

PR #223, opened 2018-09-11
Bugfix memory leak

a-szotyori:
To keep the state between activity restarts we saved the `Document` and `Photo` in `onSaveInstanceState()`. This was causing memory leaks for the following reasons:

* `onSaveInstanceState()` is called as soon as there is a chance the activity will be destroyed, meaning that it can be called multiple times without `onCreate()` or `onRestoreInstanceState()` being called afterwards.
* Data size that may be parceled is limited. To circumvent this we use the `ParcelableMemoryCache` which is a singleton and keeps references to byte arrays and bitmaps for `Documents` and `Photos` when they are parceled. During unparceling the byte arrays and bitmaps are fetched and removed from the cache.
* Due to the above leaks occurred when parceling and unparceling `Documents` and `Photos` was not balanced. In this particular case parceling happened every time the `ReviewFragmentImpl` and `AnalysisFragmentImpl` were stopped. Only rotation caused restarts were always parceling and unparceling the objects.

#### The fix:
Allow tagging of `ParcelableMemoryCache` entries. With the help of the tags clear the cache of the desired tagged entries as soon as it is safe to presume that unparceling will never happen and the cached data won't be used.

### How to test
Run one of the example apps and use the Android Profiler in AS to observe memory usage. Take pictures, send the app to the background and resume, enter and exit the review and analysis screens multiple times, etc. The memory usage should NOT show a build up of allocations.

### Ticket 
https://trello.com/c/oSvhvofV/245-crash-in-netginiandroidvision

### DIFF
--- a/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
+++ b/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
@@ -237,8 +237,8 @@ class AnalysisFragmentImpl implements AnalysisFragmentInterface {
     }
 
     private void clearParcelableMemoryCache() {
-        // Remove data from memory cache which was added when the document in the arguments
-        // was automatically parcelled when the activity has been stopped
+        // Remove data from the memory cache. The data had been added when the document in the
+        // arguments was automatically parcelled when the activity was stopped
         ParcelableMemoryCache.getInstance().removeEntriesWithTag(PARCELABLE_MEMORY_CACHE_TAG);
     }
 
--- a/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
+++ b/ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
@@ -41,6 +41,7 @@ import net.gini.android.vision.document.GiniVisionDocument;
 import net.gini.android.vision.document.GiniVisionDocumentError;
 import net.gini.android.vision.document.GiniVisionMultiPageDocument;
 import net.gini.android.vision.document.PdfDocument;
+import net.gini.android.vision.internal.camera.photo.ParcelableMemoryCache;
 import net.gini.android.vision.internal.document.DocumentRenderer;
 import net.gini.android.vision.internal.document.DocumentRendererFactory;
 import net.gini.android.vision.internal.network.AnalysisNetworkRequestResult;
@@ -66,6 +67,7 @@ import jersey.repackaged.jsr166e.CompletableFuture;
 class AnalysisFragmentImpl implements AnalysisFragmentInterface {
 
     protected static final Logger LOG = LoggerFactory.getLogger(AnalysisFragmentImpl.class);
+    private static final String PARCELABLE_MEMORY_CACHE_TAG = "ANALYSIS_FRAGMENT";
 
     private static final AnalysisFragmentListener NO_OP_LISTENER = new AnalysisFragmentListener() {
         @Override
@@ -123,6 +125,9 @@ class AnalysisFragmentImpl implements AnalysisFragmentInterface {
             final String documentAnalysisErrorMessage) {
         mFragment = fragment;
         mMultiPageDocument = asMultiPageDocument(document);
+        // Tag the documents to be able to clean up the automatically parcelled data
+        ((GiniVisionDocument) document).setParcelableMemoryCacheTag(PARCELABLE_MEMORY_CACHE_TAG);
+        mMultiPageDocument.setParcelableMemoryCacheTag(PARCELABLE_MEMORY_CACHE_TAG);
         mDocumentAnalysisErrorMessage = documentAnalysisErrorMessage;
     }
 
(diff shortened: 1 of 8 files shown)


ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -394,6 +410,7 @@ private AnalysisHint getNextHint() {
 
     void onStop() {
         mStopped = true;
+        mHandler.removeCallbacks(mHintStartRunnable);
```
a-szotyori: Remove the delayed runnable for showing the hints. This prevented the activity from being garbage collected for the duration of the `HINT_START_DELAY`.

a-szotyori:
[Crash in "net.gini.android.vision"](https://trello.com/c/oSvhvofV/245-crash-in-netginiandroidvision)

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -118,11 +121,13 @@ public void onProceedToNoExtractionsScreen(@NonNull final Document document) {
     private boolean mStopped;
     private boolean mAnalysisCompleted;
 
-
     AnalysisFragmentImpl(final FragmentImplCallback fragment, @NonNull final Document document,
             final String documentAnalysisErrorMessage) {
         mFragment = fragment;
         mMultiPageDocument = asMultiPageDocument(document);
+        // Tag the documents to be able to clean up the automatically parcelled data
+        ((GiniVisionDocument) document).setParcelableMemoryCacheTag(PARCELABLE_MEMORY_CACHE_TAG);
```
EarlOfEgo: Since `setParcelableMemoryCacheTag` is not part of `Document` the instance should be checked.

ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -265,6 +276,10 @@ public void onStart() {
         if (activity == null) {
             return;
         }
+        // Remove data from memory cache which was added when the document in the arguments
+        // was automatically parcelled when the activity has been stopped
+        ParcelableMemoryCache.getInstance().removeEntriesWithTag(PARCELABLE_MEMORY_CACHE_TAG);
```
EarlOfEgo: Instead of copy and pasting all the comments, just make a private method. 🙃

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ParcelableMemoryCache.java
```diff
@@ -61,43 +81,108 @@ public int describeContents() {
         @Override
         public void writeToParcel(final Parcel dest, final int flags) {
             dest.writeInt(tokenId);
+            dest.writeString(tag);
         }
 
         @Override
-        public boolean equals(final Object other) {
-            if (this == other) {
+        public boolean equals(final Object o) {
```
EarlOfEgo: That's not really an improvement for the name here, of course it's autogenerated.

ginivision/src/main/java/net/gini/android/vision/internal/camera/photo/ParcelableMemoryCache.java
```diff
@@ -107,15 +192,31 @@ public Token storeBitmap(@Nullable final Bitmap documentBitmap) {
...
+        }
+        LOG.debug("Store bitmap for token {} with tag {}", token, tag);
+        logCacheSizes();
         return token;
     }
 
     public void removeBitmap(@NonNull final Token token) {
         mBitmapCache.remove(token);
+        LOG.debug("Remove bitmap for token {}", token);
+        logCacheSizes();
```
EarlOfEgo: I can imagine that the logs helped during development, but I'm not sure if every method should be logged here and if it is worth to pollute the logcat so much.

ginivision/src/main/java/net/gini/android/vision/review/ReviewFragmentImpl.java
```diff
@@ -422,6 +427,8 @@ void onStop() {
     }
 
     void onSaveInstanceState(final Bundle outState) {
+        // Remove previously saved data from the memory cache
+        ParcelableMemoryCache.getInstance().removeEntriesWithTag(PARCELABLE_MEMORY_CACHE_TAG);
```
EarlOfEgo: I'm confused, why removing entries here?

ginivision/src/main/java/net/gini/android/vision/review/ReviewFragmentImpl.java
```diff
@@ -430,6 +437,13 @@ public void onDestroy() {
         if (!mNextClicked) {
             deleteUploadedDocument();
         }
+        final Activity activity = mFragment.getActivity();
+        if (activity != null && activity.isFinishing()) {
+            // Remove data from the memory cache wich was added in onSaveInstanceState() and also
```
EarlOfEgo: >which was added in onSaveInstanceState()
What was added in `onSaveInstanceState`? (also there is a typo in which)

ginivision/src/main/java/net/gini/android/vision/review/multipage/previews/PreviewFragment.java
```diff
@@ -209,6 +218,14 @@ public void onStop() {
         mStopped = true;
     }
 
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        // Remove data from memory cache which was added when the document in the arguments
+        // was automatically parcelled when the activity has been stopped
+        ParcelableMemoryCache.getInstance().removeEntriesWithTag(PARCELABLE_MEMORY_CACHE_TAG);
```
EarlOfEgo: Just as above, a method with a describing name would be nicer to read.

ginivision/src/main/java/net/gini/android/vision/review/ReviewFragmentImpl.java
a-szotyori: `onSaveInstanceState()` can be called multiple times without an `onCreate()` or `onRestoreInstanceState()` call afterwards. Clearing the cache here makes sure that there are no "dangling" entries in the cache.

PR #230, opened 2018-11-02
MultiPageReviewFragment doesn't finish its hosting Activity directly anymore

a-szotyori:
Using a single activity to host all the Component API fragments revealed an oversight in the `MultiPageReviewFragment`. It was finishing its host activity directly when the user tapped on the add more pages button or deleted the last page. This prevented the client app from being able to go back to the camera fragment in the same activity.

### How to test
Run our example apps.


a-szotyori:
[Fix MultiPageReviewFragment finishing its host Activity](https://trello.com/c/kygiHrbJ/353-fix-multipagereviewfragment-finishing-its-host-activity)

PR #238, opened 2019-01-16
Network library implementation using the accounting API

a-szotyori:
**Disclaimer:** Won't compile until the v2.3.0 of the Gini API SDK is released. Sorry about that 🙈.

Created a new network library module for the accounting API. It uploads the image to the accounting API in the review screen without the user rotation information. The getting the extractions is done in the analysis screen.

If it's used with the multi-page feature it will generate an error when the `analyze()` method is called.

Extended the example apps to allow switching between the default and the accounting networking library.

### How to test
Run the example apps and use the accounting gini api type.


ginivision-accounting-network/src/main/java/net/gini/android/vision/accounting/network/GiniVisionAccountingNetworkService.java
```diff
@@ -0,0 +1,559 @@
...
+                                    giniApiDocumentId);
+                            callback.cancelled();
+                        }
+                        return null;
+                    }
+                }, Task.UI_THREAD_EXECUTOR);
+        return new NoOpCancellationToken();
+    }
+
+    @SuppressWarnings("PMD.ExcessiveMethodLength")
```
kikettas: Seeing this I guess it's impossible to reduce this method length, isn't it?

ginivision-accounting-network/src/main/java/net/gini/android/vision/accounting/network/GiniVisionAccountingNetworkService.java
```diff
@@ -0,0 +1,559 @@
...
+                                    LOG.debug(
+                                            "Document analysis cancelled after polling for document {}",
+                                            giniApiDocumentIdRotationMap);
+                                    return Task.cancelled();
+                                }
+                                final net.gini.android.models.Document giniApiDocument =
+                                        task.getResult();
+                                if (task.isCancelled()) {
+                                    LOG.debug(
+                                            "Composite document polling cancelled for document {}",
```
kikettas: ~Composite~

ginivision-accounting-network/src/main/java/net/gini/android/vision/accounting/network/model/BoxMapper.java
```diff
@@ -0,0 +1,54 @@
+package net.gini.android.vision.accounting.network.model;
+
+import android.support.annotation.Nullable;
+
+import net.gini.android.models.Box;
+import net.gini.android.vision.network.model.GiniVisionBox;
+
+/**
+ * Created by Alpar Szotyori on 30.01.2018.
```
kikettas: Almost a year since we implemented the networking plugin...

ginivision-accounting-network/src/main/java/net/gini/android/vision/accounting/network/GiniVisionAccountingNetworkService.java
a-szotyori: Well, given enough time and money, it's doable I guess 😛.

kikettas: Fair point 👍

PR #242, opened 2019-02-08
Feature default app dialog

a-szotyori:
When GVL is launched with an external file (aka "open with") we now check whether the app is set as default for that file type. If the app is the default, then we show an alert informing the user about this and telling them how to revert this. The user can either continue or be directed to the app's screen in the Android Settings app.

For images the dialog is shown in the review screens and for pdfs in the analysis screen.

*Note:* Showing an alert was only possible in the camera screen before. I had to touch a lot of files to allow the other screen to show dialogs, too, with minimal code duplication.

### How to test
Open a pdf or an image with one of the example apps and use the `ALWAYS` option. The dialog should be shown.


ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -445,16 +449,28 @@ ProgressBar getProgressActivity() {
...
-        if (mFragment.getActivity() == null) {
+        final Activity activity = mFragment.getActivity();
+        if (activity == null) {
             return;
         }
-        if (mDocumentAnalysisErrorMessage != null) {
-            showError(mDocumentAnalysisErrorMessage,
-                    mFragment.getActivity().getString(R.string.gv_document_analysis_error_retry),
-                    new View.OnClickListener() {
+        if (MimeType.APPLICATION_PDF.asString().equals(mMultiPageDocument.getMimeType())) {
```
kikettas: This 🔺 cannot be avoided, can it?

PR #244, opened 2019-02-18
Toggleable flash

sz-alpar:
Added a flash toggle button to the camera screen. The button is hidden by default. Flash is on by default and if the flash button is enabled the user can toggle between flash on and off. If flash is not available the button is always hidden.

### How to test
Run the example apps on phones and tablets (with and without flash).


a-szotyori:
Created with my personal account by accident 🙈. Will recreate it with the company one.

PR #245, opened 2019-02-18
Toggleable flash

a-szotyori:
Added a flash toggle button to the camera screen. The button is hidden by default. Flash is on by default and if the flash button is enabled the user can toggle between flash on and off. If flash is not available the button is always hidden.

### How to test
Run the example apps on phones and tablets (with and without flash).

### Some screenshots

phone-16-9
phone-large
phone-normal
phone-small
tablet-landscape
tablet-portrait


ginivision/src/main/java/net/gini/android/vision/internal/camera/view/FlashButtonHelper.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @exclude
+ */
+public final class FlashButtonHelper {
+
+    /**
+     * @exclude
+     */
+    public enum FLASH_BUTTON_POSITION {
```
kikettas: This underscore is kinda weird for the enum name, wouldn't it be better to use `FlashButtonPosition`?

a-szotyori: Oh, yeah, you're right!

PR #246, opened 2019-03-01
Fix test execution

a-szotyori:
Migrated test dependencies from support library to androidx.

Found a way to run instrumented tests on a specific device or emulator using the Android Gradle plugin's `connectedAndroidTest` task. This allows us to get rid of our own gradle task and shell script used to enable targeting specific devices 🎉.

Fixed an issue in the Analysis Screen where error messages from the Review Screen were ignored and analysis was started. This happens only when multi-page is not enabled and was introduced in version 3.5.0. It's not critical as it uploads the document again without showing the user the error from the Review Screen. If an error occurs with this upload the error is shown to the user.

### How to test
Launch an emulator and run `$ ./gradlew ginivision:connectedAndroidTest`. The test results can be viewed with `ginivision/build/reports/androidTests/connected/index.html`. Alternatively run the tests using Android Studio.


ginivision/src/main/java/net/gini/android/vision/analysis/AnalysisFragmentImpl.java
```diff
@@ -454,23 +454,27 @@ private void analyzeDocument() {
...
-                                        new View.OnClickListener() {
-                                            @Override
-                                            public void onClick(final View v) {
-                                                doAnalyzeDocument();
-                                            }
-                                        });
-                            } else {
-                                doAnalyzeDocument();
-                            }
+                            showErrorIfAvailableAndAnalyzeDocument(activity);
```
kikettas: 👍

PR #254, opened 2019-07-01
Stores images in a subfolder to only affect own files

a-szotyori:
We omitted putting GVL images into a subfolder and only working with that to not affect host app's files. This PR remedies that by storing GVL images in the `gv-images` subfolder.

### How to test
Run one of the example apps using Android Studio. Navigate using the Device File Explorer to the apps private files in `data/data/net.gini.android.vision.*apiexample/files`. Upload a file there and verify that after analysing documents the `gv-images` folder is removed, but all other files remain.

### Ticket 
[PIA-298](https://ginis.atlassian.net/browse/PIA-298)


skaphander:
Does `ginivision/src/main/java/net/gini/android/vision/internal/storage/ImageDiskStore.java` have a dedicated test? Why is the presenter test adapted?

a-szotyori:
Now the `ImageDiskStore` has dedicated tests. 🙈

PR #260, opened 2019-12-04
Added camera permission check before opening the camera

a-szotyori:
On some devices (like the Huawei P8 Lite) relying on the camera error to detect that camera permission was not granted is not reliable.

I added a camera permission check before using the camera and if permission wasn't granted I show the no permission views.

### Ticket 
[PIA-398](https://ginis.atlassian.net/browse/PIA-398)

### DIFF
--- a/ginivision/src/androidTest/java/net/gini/android/vision/test/PermissionsHelper.java
+++ b/ginivision/src/androidTest/java/net/gini/android/vision/test/PermissionsHelper.java
@@ -50,6 +50,20 @@ public class PermissionsHelper {
         }
     }
 
+    public static void revokeCameraPermission() throws InterruptedException {
+        revokePermission(Manifest.permission.CAMERA, getApplicationContext().getPackageName());
+    }
+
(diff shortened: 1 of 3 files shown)


ginivision/src/androidTest/java/net/gini/android/vision/test/PermissionsHelper.java
```diff
@@ -50,6 +50,20 @@ private static void grantPermission(final String permission, final String packag
...
+        revokePermission(Manifest.permission.CAMERA, getApplicationContext().getPackageName());
+    }
+
+    private static void revokePermission(final String permission, final String packageName) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            if (!hasPermission(permission, packageName)) {
+                return;
+            }
+            final String command = "pm revoke " + packageName + " " + permission;
+            executeShellCommand(command);
```
maciejtrybilo: 😅

PR #298, opened 2020-10-16
Make locale customizable

ThiesR:
Added the ability to customize the locale that is used for translating string resources. An abstract super class for activities was introduced to assure that the custom locale will be applied. The super class also provides a helper method for translating the activitie's title.

### How to test
Set a custom locale when initializing GiniVision, that is different to the locale of the device. The GiniVision screens should use the custom locale.


a-szotyori:
Hi, thanks for the PR! Unfortunately we can't merge it. We adhere to Android standards as much as possible. This also includes localization. We see no advantage in allowing clients to set a locale different from what the system uses.

You may fork our repository and implement your changes there. Of course you will need to maintain and fix it in case we introduce breaking changes related to your modifications.

We are sorry we can't fulfill your request as we have many other clients and need to adhere to Android standards.

PR #316, opened 2021-06-18
Feature/pia 647 qr code help

a-szotyori:
Please review after #315.

* Extract hint popup logic into a dedicated class.
* Show QR code scanner hint. It is shown after the upload hint unless the upload hint was already shown or upload is not enabled.

Will update the customization guide on Monday.


zladzeyka (approved):
Looks good,👍 😊 ,
I'll test it on Monday and came back to you, @a-szotyori

PR #322, opened 2021-08-11
Remove jcenter and update gini-android-sdk to 2.9.0

a-szotyori:
Had to update gini-android-sdk to use Volley 1.2.0 which is the first
version available on maven central.

### Ticket 
https://ginis.atlassian.net/browse/PIA-1190

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,6 @@ buildscript {
 
     repositories {
         google()
-        jcenter()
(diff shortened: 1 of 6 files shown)


a-szotyori:
The build failed, but it was fixed in #323 so please ignore the failure.

zladzeyka (approved):

PR #323, opened 2021-08-11
Pia 1192 update to gradle7 publish plugin

a-szotyori:
Please review after #322.

Migrate to gradle 7 and replace the old maven plugin with the new `maven-publish` plugin for publishing releases to our own maven repository.

Also update the Android Gradle Plugin to the latest version (7.0.0) and use the required Java 11 to run gradle.

For creating javadoc we use AGP 4.2.2 and Java 8. With Java 11 (which is required for AGP 7.0.0) the Javadoc plugin of Gradle is broken and no javadoc is created.

### Ticket 
https://ginis.atlassian.net/browse/PIA-1192


zladzeyka (approved):
👍🏻

PR #326, opened 2021-08-17
Target sdk version 31 (Android 12) and set `android:exported` attribute

a-szotyori:
When targeting Android 12 activities with intent filters require
the `android:exported` attribute:
https://developer.android.com/about/versions/12/behavior-changes-12#exported

### Ticket 
https://ginis.atlassian.net/browse/PIA-1598


zladzeyka (approved):

zladzeyka:
@a-szotyori some test is failing

